PR #16, opened 2016-06-22
default gradle tasks instead of calling gradle tasks cmd

godrei:


scanners/android/android.go
```diff
@@ -243,10 +213,7 @@ func (scanner *Scanner) Options() (models.OptionModel, error) {
 		logger.InfofSection("Inspecting gradle file: %s", gradleFile)
```
viktorbenei: remove this as well, I guess

viktorbenei:
LGTM 👍

PR #38, opened 2016-08-03
xcodeproj and xcworkspace should be a dir

godrei:


godrei:
write test for:
- if xcodeproj/xcworkspace is symlink
- if xcodeproj/xcworkspace is file

PR #46, opened 2016-09-20
Warnings

godrei:


viktorbenei:
ohh the irony :D ("do not fail if detectors fail" fails :P)

PR #74, opened 2017-02-23
Errors

godrei:


cli/config.go
```diff
@@ -49,6 +50,16 @@ var configCommand = cli.Command{
 	},
 }
 
+func writeOutput(outputDir string, scanResult models.ScanResultModel, format output.Format) (string, error) {
```
trapacska: writeOutput seems too mainstream, it could be for example: writeScanResultOutput

cli/config.go
```diff
@@ -49,6 +50,16 @@ var configCommand = cli.Command{
...
+	pth := path.Join(outputDir, "result")
+	outputPth, err := output.WriteToFile(scanResult, format, pth)
```
trapacska: can be a return here

models/model_methods.go
```diff
@@ -0,0 +1,126 @@
...
+			PullRequestSourceBranch: "*",
+			WorkflowID:              primaryWorkflowID,
+		},
+	}
+
+	app := bitriseModels.AppModel{
+		Environments: appEnvs,
+	}
+
+	bitriseData := bitriseModels.BitriseDataModel{
```
trapacska: maybe a return could be here

models/model_methods.go
```diff
@@ -0,0 +1,126 @@
...
```
trapacska: a return can be here also

PR #81, opened 2017-04-04
Added change-workdir in android workflows

trapacska:


scanners/android/android.go
```diff
@@ -296,6 +312,13 @@ func (scanner *Scanner) DefaultConfigs() (models.BitriseConfigMap, error) {
 	// GitClone
 	stepList = append(stepList, steps.GitCloneStepListItem())
 
+	// ChangeWorkdir
+	cwdInputs := []envmanModels.EnvironmentItemModel{
+		envmanModels.EnvironmentItemModel{"path": "$" + projectDirEnvKey},
```
godrei: use `projectDirKey` instead of `path`

PR #87, opened 2017-05-30
error if Android `local.properties` is committed into the repo - completed

szokezsolt:
Scanner (or somewhere else? e.g. in Gradle Runner?): warning if Android `local.properties` is committed into the repo

If the local.properties file is committed into the repo you'll get an error for any Android/Gradle build like:

```
* What went wrong:
A problem occurred configuring project ':app'.
> The SDK directory '/Users/rodrigosimoesrosa/Library/Android/sdk' does not exist.
```

Note the directory path -> it's the path __on the user's Mac!__

### DIFF
--- a/scanners/android/android.go
+++ b/scanners/android/android.go
@@ -108,19 +108,21 @@ func (scanner *Scanner) ExcludedScannerNames() []string {
 
 // Options ...
 func (scanner *Scanner) Options() (models.OptionModel, models.Warnings, error) {
+	warnings := models.Warnings{}
 
 	// Search for local.properties file
 	for _, filePath := range scanner.FileList {
 		if strings.Contains(filePath, "local.properties") {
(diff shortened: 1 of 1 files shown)


scanners/android/android.go
```diff
@@ -107,6 +108,15 @@ func (scanner *Scanner) ExcludedScannerNames() []string {
 
 // Options ...
 func (scanner *Scanner) Options() (models.OptionModel, models.Warnings, error) {
+
+	// Search for local.properties file
+	for _, filePath := range scanner.FileList {
+		if strings.Contains(filePath, "local.properties") {
+			log.Warnft(`the local.properties file should not be committed into the repository. The location of the file is:
```
trapacska: Question: Does it should be in the `warnings := models.Warnings{}` model, as we want to display the warning message on the UI as well, not only in the scanner's log?

PR #103, opened 2017-11-08
Add export_method in xcode-archive step for ios scanner

kdobmayer:


scanners/ios/utility.go
```diff
@@ -42,6 +42,21 @@ const (
...
+	ExportMethodInputEnvKey = "BITRISE_EXPORT_METHOD"
+	// ExportMethodInputTitle ...
+	ExportMethodInputTitle = "Export method"
+)
+
+// IosExportMethods ...
+var IosExportMethods = []string{"app-store", "ad-hoc", "enterprise", "development"}
+
+// MacExportMethods ...
+var MacExportMethods = []string{"app-store", "developer-id", "none", "development"}
```
godrei: `none` should be the last option

scanners/ios/utility.go
```diff
@@ -42,6 +42,21 @@ const (
 )
 
 const (
+	// ExportMethodInputKey ...
+	ExportMethodInputKey = "export_method"
+	// ExportMethodInputEnvKey ...
+	ExportMethodInputEnvKey = "BITRISE_EXPORT_METHOD"
+	// ExportMethodInputTitle ...
+	ExportMethodInputTitle = "Export method"
```
godrei: define separate title's for mac and ios platforms.

For ios:

```
ipa export method
```

For mac:

```
Application export method  
NOTE: `none` means:  Export a copy of the application without re-signing.
```

PR #115, opened 2018-05-09
New Android scanner

trapacska:


models/workflowbuilder.go
```diff
@@ -3,7 +3,8 @@ package models
 import bitriseModels "github.com/bitrise-io/bitrise/models"
 
 type workflowBuilderModel struct {
-	Steps []bitriseModels.StepListItemModel
+	Steps                []bitriseModels.StepListItemModel
+	Description, Summary string
```
godrei: Do we use `Summary` ?

scanners/android/android.go
```diff
@@ -29,38 +31,44 @@ func (*Scanner) ExcludedScannerNames() []string {
 }
```
BirmacherAkos: return nil instead of an empty array?

scanners/android/utility.go
```diff
@@ -1,65 +1,115 @@
...
+	GradlewPathInputEnvKey = "GRADLEW_PATH"
+	GradlewPathInputTitle  = "Gradlew file path"
 )
 
-// CollectRootBuildGradleFiles - Collects the most root (mint path depth) build.gradle files
-// May the searchDir contains multiple android projects, this case it return multiple builde.gradle path
-// searchDir/android-project1/build.gradle, searchDir/android-project2/build.gradle, ...
-func CollectRootBuildGradleFiles(searchDir string) ([]string, error) {
-	fileList, err := utility.ListPathInDirSortedByComponents(searchDir, true)
+func walk(src string, fn func(path string, info os.FileInfo) error) error {
```
godrei: isn't it the filepath.Walk fn: https://golang.org/pkg/path/filepath/#Walk ?

scanners/android/android.go
```diff
@@ -29,38 +31,44 @@ func (*Scanner) ExcludedScannerNames() []string {
...
 
 // DetectPlatform ...
-func (scanner *Scanner) DetectPlatform(searchDir string) (bool, error) {
+func (scanner *Scanner) DetectPlatform(searchDir string) (b bool, err error) {
```
BirmacherAkos: Do not use named return value if you don't need it.

scanners/android/utility.go
```diff
@@ -69,122 +119,328 @@ The location of the file is: %s`, localPropertiesPth)
...
+	}
+
+	return nil
 }
 
-// GenerateOptions ...
-func GenerateOptions(searchDir string) (models.OptionModel, models.Warnings, error) {
-	warnings := models.Warnings{}
+func installMissingAndroidTools(srcRoot, androidHome string) error {
+	gradlewPath := filepath.Join(srcRoot, "gradlew")
```
godrei: srcRoot can be empty? since instead of: `gradlew [command]` you would need to run: `./gradlew [command]`

_tests/integration/android_test.go
```diff
@@ -111,22 +115,77 @@ var sampleAppsAndroidSDK22SubdirVersions = []interface{}{
...
+            env_key: MODULE
+            value_map:
+              app:
+                title: Variant for building
+                env_key: BUILD_VARIANT
+                value_map:
+                  AndroidTest:
+                    title: Variant for testing
+                    env_key: TEST_VARIANT
+                    value_map:
```
godrei: is variant reqquired?

_tests/integration/android_test.go
```diff
@@ -201,22 +316,77 @@ var sampleAppsAndroid22Versions = []interface{}{
...
+	steps.AndroidUnitTestVersion,
 	steps.DeployToBitriseIoVersion,
 	steps.CachePushVersion,
 }
 
 var sampleAppsAndroid22ResultYML = fmt.Sprintf(`options:
   android:
-    title: Path to the gradle file to use
-    env_key: GRADLE_BUILD_FILE_PATH
+    title: The root directory of an Android project
```
godrei: it seems that this test tests the same as `sampleAppsAndroidSDK22SubdirResultYML` at least the generated config seems to be the same, do we need this test as well?

_tests/integration/android_test.go
```diff
@@ -281,22 +507,77 @@ var androidNonExecutableGradlewVersions = []interface{}{
...
+	steps.AndroidUnitTestVersion,
 	steps.DeployToBitriseIoVersion,
 	steps.CachePushVersion,
 }
 
 var androidNonExecutableGradlewResultYML = fmt.Sprintf(`options:
   android:
-    title: Path to the gradle file to use
-    env_key: GRADLE_BUILD_FILE_PATH
+    title: The root directory of an Android project
```
godrei: same question, do we need this test as well?

_tests/integration/reactnative_test.go
```diff
@@ -76,45 +76,213 @@ var sampleAppsReactNativeSubdirVersions = []interface{}{
...
     value_map:
-      project/android/build.gradle:
+      project/android:
         title: Gradlew file path
         env_key: GRADLEW_PATH
         value_map:
           project/android/gradlew:
-            title: Project (or Workspace) path
-            env_key: BITRISE_PROJECT_PATH
+            title: Module
```
godrei: do we want to update the react scanner as well?

scanners/android/android.go
```diff
@@ -29,38 +31,44 @@ func (*Scanner) ExcludedScannerNames() []string {
...
 func (*Scanner) DefaultOptions() models.OptionModel {
-	gradleFileOption := models.NewOption(GradleFileInputTitle, GradleFileInputEnvKey)
+	projectLocationOption := models.NewOption(ProjectLocationInputTitle, ProjectLocationInputEnvKey)
 
-	gradlewPthOption := models.NewOption(GradlewPathInputTitle, GradlewPathInputEnvKey)
-	gradleFileOption.AddOption("_", gradlewPthOption)
+	moduleOption := models.NewOption(ModuleInputTitle, ModuleInputEnvKey)
+	projectLocationOption.AddOption("_", moduleOption)
+
+	testVariantOption := models.NewOption(TestVariantInputTitle, TestVariantInputEnvKey)
```
godrei: do we need to collect the variants for the default config as well? it is a not required input of the steps and we should create a less user question as possible.

scanners/reactnative/reactnative.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/bitrise-core/bitrise-init/steps"
 	"github.com/bitrise-core/bitrise-init/utility"
 	envmanModels "github.com/bitrise-io/envman/models"
+	"github.com/bitrise-io/go-utils/command"
```
godrei: i would be much happy if we could avoid touching this scanner for now

_tests/integration/android_test.go
trapacska: It is not the same. `sampleAppsAndroidSDK22SubdirResultYML` has 'src' as project location option, and this one has `.` as project location

scanners/android/utility.go
godrei:
```
func walk2(src string, fn func(path string, info os.FileInfo) error) error {
	return filepath.Walk(src, func(path string, info os.FileInfo, err error) error {
		if err != nil {
			return err
		}
		if path == src {
			return nil
		}
		return fn(path, info)
	})
}
```

scanners/android/utility.go
```diff
@@ -69,122 +111,330 @@ The location of the file is: %s`, localPropertiesPth)
...
+		}
+	}
+
+	return nil
 }
 
-// GenerateOptions ...
-func GenerateOptions(searchDir string) (models.OptionModel, models.Warnings, error) {
-	warnings := models.Warnings{}
+func installMissingAndroidTools(srcRoot, androidHome string) error {
```
godrei: could not we make this function exported in the install missing android tools step? i would be much happier if we would not duplicate these long logics...

PR #119, opened 2018-05-22
Reactnavite-expo scanner

BirmacherAkos:


scanners/reactnative-expo/reactnative-expo.go
```diff
@@ -0,0 +1,706 @@
...
+
+	if !dependencyFound {
+		return false, nil
+	}
+
+	log.TPrintf("%d package.json file detected", len(packageJSONPths))
+
+	log.TInfof("Filter relevant package.json files")
+
+	fmt.Printf("Returning true for expo")
```
godrei: what is this log for?

scanners/reactnative-expo/reactnative-expo.go
```diff
@@ -0,0 +1,706 @@
...
+	iosScanner := ios.NewScanner()
+	androidScanner := android.NewScanner()
+	for _, packageJSONPth := range packageJSONPths {
```
godrei: why do we check that there are no native projects?

scanners/reactnative-expo/reactnative-expo.go
```diff
@@ -0,0 +1,706 @@
...
+		} else if exist {
+			if detected, err := androidScanner.DetectPlatform(scanner.searchDir); err != nil {
+				return false, err
+			} else if detected {
+				return false, nil
+			}
+		}
+	}
+
+	if err := ensureNodeModules(); err != nil {
```
godrei: may we could pre-validate if there is an eject npm task or not (the npm tasks are listen in the package.json file)

scanners/reactnative-expo/reactnative-expo.go
```diff
@@ -0,0 +1,706 @@
...
+		log.Errorf("ERROR DURING INSTALLING NPM: %s", err)
+		return false, nil
+	}
+
+	if err := ejectProject(searchDir); err != nil {
```
godrei: may we could pre-validate if there is an eject npm task or not (the npm tasks are listen in the package.json file)

scanners/reactnative-expo/reactnative-expo.go
```diff
@@ -0,0 +1,706 @@
...
+		if iosProjectDetected || androidProjectDetected {
+			relevantPackageJSONPths = append(relevantPackageJSONPths, packageJSONPth)
+		} else {
+			log.TWarnf("no ios nor android project found, skipping package.json file")
+		}
+	}
+
+	if len(relevantPackageJSONPths) == 0 {
+		return false, nil
+	}
```
godrei: if more then 1 package.json file found list them and log using warning level which one will we use

scanners/reactnative-expo/reactnative-expo.go
```diff
@@ -0,0 +1,674 @@
...
+	yaml "gopkg.in/yaml.v2"
+)
+
+// Name ...
+const Name = "react-native-expo"
+
+const (
+	workDirInputKey = "workdir"
+
+	gradleFileInputKey    = "gradle_file"
```
godrei: unused

scanners/reactnative-expo/reactnative-expo.go
```diff
@@ -0,0 +1,674 @@
...
+	gradleFileInputEnvKey = "GRADLE_BUILD_FILE_PATH"
```
godrei: this should be defined in the android scanner

scanners/reactnative-expo/reactnative-expo.go
```diff
@@ -0,0 +1,674 @@
...
+		}
+
+	}
+
+	return rootOption, warnings, nil
+}
+
+// DefaultOptions ...
+func (Scanner) DefaultOptions() models.OptionModel {
+	gradleFileOption := models.NewOption(gradleFileInputTitle, gradleFileInputEnvKey)
```
godrei: could not we use the ios and android scanner's default options?

scanners/reactnative-expo/reactnative-expo.go
```diff
@@ -0,0 +1,674 @@
...
+			lineModification.lastTime = time.Now()
+			LastLineModification.Store(lineModification)
+		}
+	}()
+
+	ch := make(chan bool)
+	stopWaiting := false
+	go func(ch *chan bool) {
+		for {
+			go checkLastLineTime(ch)
```
godrei: why do we do this in a new go routine?

scanners/reactnative-expo/reactnative-expo.go
BirmacherAkos: I think it would be better if we duplicate this one because it could be difficult to modify the Android or iOS scanner if we use them everywhere.

godrei: that is true, but we want to use the same options here

scanners/reactnative-expo/reactnative-expo.go
```diff
@@ -0,0 +1,665 @@
...
+	yaml "gopkg.in/yaml.v2"
+)
+
+// Name ...
+const Name = "react-native-expo"
+
+const (
+	workDirInputKey = "workdir"
+
+	gradleFileInputEnvKey = "GRADLE_BUILD_FILE_PATH"
```
godrei: use the same options and configs as the ios and android scanner.

in the new android scanner we do not use `GRADLE_BUILD_FILE_PATH `

scanners/reactnative-expo/reactnative-expo.go
godrei: other than this you add this option, but is not used in any steo

scanners/reactnative-expo/reactnative-expo.go
```diff
@@ -0,0 +1,665 @@
...
+	cmd.SetStdout(os.Stdout)
+	cmd.SetStderr(os.Stderr)
+	cmd.SetDir(cmdDir)
+	return cmd.Run()
+}
+
+func ejectProject(pth string, cmdDir string) error {
+	log.Infof("Eject project")
+
+	lineModification := LineModification{lastTime: time.Now(), answerCount: 0}
```
godrei: this could be less complicated:

```
	cmd := exec.Command("bash", "./test.sh")

	stdin, err := cmd.StdinPipe()
	if err != nil {
		panic(err)
	}

	stdoutReader, err := cmd.StdoutPipe()
	if err != nil {
		panic(err)
	}

	scanner := bufio.NewScanner(stdoutReader)
	var t *time.Timer

	go func() {
		for scanner.Scan() {
			if t != nil {
				t.Stop()
			}

			line := scanner.Text()
			fmt.Println(line)

			t = time.AfterFunc(5*time.Second, func() {
				if _, err := io.WriteString(stdin, "test\n"); err != nil {
					panic(err)
				}
			})
		}
		if err := scanner.Err(); err != nil {
			panic(err)
		}
	}()

	if err := cmd.Run(); err != nil {
		panic(err)
	}
```

scanners/reactnative-expo/utility.go
```diff
@@ -0,0 +1,61 @@
...
+	name += "-config"
+	return name
+}
+
+func defaultConfigName() string {
+	return "default-react-native-expo-config"
+}
+
+// FindDependency ...
+func FindDependency(filePath string, dependency string, parentKey string) (bool, error) {
```
godrei: we have cordova. ParsePackagesJSON for this functionality

_tests/integration/manual_config_test.go
```diff
@@ -343,13 +343,13 @@ var customConfigResultYML = fmt.Sprintf(`options:
                     env_key: BITRISE_EXPORT_METHOD
                     value_map:
                       ad-hoc:
-                        config: default-react-native-expo-config
+                        config: default-ios-config
```
godrei: default-ios-config ? this should be default-react-native-expo-config

scanners/reactnative-expo/utility.go
```diff
@@ -0,0 +1,67 @@
...
+	packages, err := cordova.ParsePackagesJSON(filePath)
+	if err != nil {
+		return false, err
+	}
+
+	log.TPrintf("Searching for %s", dep)
+
+	dependencyFound := false
+	for dependency := range packages.Dependencies {
+		if strings.Contains(dependency, dep) {
```
godrei: we need exact macth

scanners/reactnative-expo/utility.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if !dependencyFound {
+		return false, nil
+	}
+
+	log.TPrintf("Searching for %s", scrt)
+
+	dependencyFound = false
+	for script := range packages.Scripts {
+		if strings.Contains(script, scrt) {
```
godrei: we need exact match

_tests/integration/manual_config_test.go
```diff
@@ -302,6 +322,34 @@ var customConfigResultYML = fmt.Sprintf(`options:
                         config: default-react-native-config
                       enterprise:
                         config: default-react-native-config
+  react-native-expo:
+    title: The root directory of an Android project
+    env_key: PROJECT_LOCATION
+    value_map:
+      _:
```
trapacska: missing `GRADLEW_PATH` - required for install missing android tools

PR #130, opened 2018-12-17
Flutter

trapacska:


scanners/flutter/flutter.go
```diff
@@ -0,0 +1,261 @@
...
+
+	return true, nil
+}
+
+// ExcludedScannerNames ...
+func (Scanner) ExcludedScannerNames() []string {
+	return []string{
+		string(ios.XcodeProjectTypeIOS),
+		string(ios.XcodeProjectTypeMacOS),
+		cordova.ScannerName,
```
godrei: why does it exclude cordova? if it does why does not exclude it ionic?

scanners/flutter/flutter.go
```diff
@@ -0,0 +1,261 @@
...
+		for xcodeWorkspacePath, schemes := range project.xcodeProjectPaths {
+			schemeOption := models.NewOption(ios.SchemeInputTitle, ios.SchemeInputEnvKey)
+			projectPathOption.AddOption(xcodeWorkspacePath, schemeOption)
+
+			for _, scheme := range schemes {
+				exportMethodOption := models.NewOption(ios.IosExportMethodInputTitle, ios.ExportMethodInputEnvKey)
+				schemeOption.AddOption(scheme, exportMethodOption)
+
+				for _, exportMethod := range ios.IosExportMethods {
+					configOption := models.NewConfigOption(defaultConfigName)
```
godrei: is it intended: using the same config name for default and scanned workflows?

godrei: i see so the default and scanned configs are the same, so they share a name, this case my i would not add the default suffix to the workflow name

PR #135, opened 2019-01-15
Set variant to be empty

trapacska:


scanners/android/android.go
```diff
@@ -79,7 +79,7 @@ func (scanner *Scanner) DefaultOptions() models.OptionModel {
 
 	projectLocationOption.AddOption("_", moduleOption)
 	moduleOption.AddOption("_", variantOption)
-	variantOption.AddOption("_", configOption)
+	variantOption.AddConfig("", configOption)
```
godrei: in the case of `DefaultOptions` should not `_` remain?

trapacska: I think we shall not force it to be filled as it is not a required field.

PR #140, opened 2019-02-22
Gradle kts support

lszucs:


scanners/android/android.go
```diff
@@ -40,7 +40,18 @@ func (scanner *Scanner) DetectPlatform(searchDir string) (_ bool, err error) {
 		return false, fmt.Errorf("failed to search for build.gradle files, error: %s", err)
 	}
 
-	return len(scanner.ProjectRoots) > 0, err
+	isRecognized := len(scanner.ProjectRoots) > 0
```
trapacska: I think this is not necessary, can be moved to the if simply, also then the return can be something like 
`return len(scanner.ProjectRoots) > 0, err`

lszucs: It is not immediately obvious, that `len(scanner.ProjectRoots) > 0` means, that the scanner successfully recognized a project. One must read and interpret the surrounding code. Using the properly named variable, this is not an issue.

scanners/android/android.go
```diff
@@ -40,7 +40,18 @@ func (scanner *Scanner) DetectPlatform(searchDir string) (_ bool, err error) {
...
+
+	if !isRecognized {
```
trapacska: This way kotlin projects will be collected only if there was no android projects at all. It is up to the user if he mixes his project base languages. Please collect both android and kotlin based project paths.

trapacska: This way when the scanner asks the user which repos he want to select from, he will be able to select from both type.

scanners/flutter/flutter.go
```diff
@@ -179,6 +179,14 @@ projects:
 			proj.hasAndroidProject = true
 		}
 
+		// todo: duplicate code; factor into detectAndroidProjectPath function?
```
lpusok: This is somewhat similar: https://github.com/bitrise-tools/go-android/blob/master/gradle/project.go

lszucs: Thanks, will discuss offline if it is usable here. TODO comment will be removed before merge.

PR #144, opened 2019-03-22
Add possible ios and android app icons as options

lpusok:


cli/icon.go
```diff
@@ -0,0 +1,50 @@
...
+	for iconID, iconPath := range icons {
+		if err := copyFile(iconPath, filepath.Join(outputDir, iconID)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func copyFile(src string, dst string) (err error) {
+	from, err := os.Open(src)
```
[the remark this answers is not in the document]
lpusok: It is good to know, that in this case all contents will be read from the file to memory.

scanners/android/icon/icon.go
```diff
@@ -0,0 +1,94 @@
...
+		if app == nil {
+			log.TPrintf("Key application not found in manifest file")
+			return "", nil
+		}
+		ic := app.SelectAttr("android:icon")
+		if ic == nil {
+			log.TPrintf("Attribute not found in manifest file")
+			return "", nil
+		}
+		iconName = strings.TrimPrefix(ic.Value, `@mipmap/`)
```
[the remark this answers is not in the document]
lpusok: It still does occur in the wild, but officially mipmap should be used: https://developer.android.com/guide/topics/resources/providing-resources

scanners/ios/ios.go
```diff
@@ -40,15 +41,15 @@ func (Scanner) ExcludedScannerNames() []string {
...
+	options, configDescriptors, icons, warnings, err := GenerateOptions(XcodeProjectTypeIOS, scanner.SearchDir, scanner.ExcludeAppIcon)
 	if err != nil {
-		return models.OptionNode{}, warnings, err
+		return models.OptionNode{}, warnings, models.Icons{}, err
 	}
 
 	scanner.ConfigDescriptors = configDescriptors
 
-	return options, warnings, nil
+	return options, warnings, icons, nil
```
BirmacherAkos: Should we return all the icons?
Maybe it would be enough to return only one with the highest resolution.

scanners/ios/icon/icon.go
```diff
@@ -0,0 +1,142 @@
...
+	"strings"
+
+	"github.com/bitrise-core/bitrise-init/models"
+	"github.com/bitrise-core/bitrise-init/utility"
+	"github.com/bitrise-io/go-utils/log"
+)
+
+// LookupPossibleMatches returns possible ios app icons,
+// in a map with key of a id (sha256 hash converted to string), value of icon path
+func LookupPossibleMatches(searchPath string, basepath string) (models.Icons, error) {
```
godrei: It seems the given target's `buildPhases` contains which .xcassets will be copied into the .app bundle (a `PBXResourcesBuildPhase` entry, which `files` node contains .xcassets).
The target's `ASSETCATALOG_COMPILER_APPICON_NAME ` describes which asset is the app icon from the copied `.xcassets`.

The strange thing is that multiple copied `.xcassets` can contain app icon assets with the same name, in this case Xcode is buggy a bit and you can not select which one you want to use (there is a dropdown in xcode and you can select which catalogs which assets you want to use, but you can not switch between app icon assets with the same name in different assets).

lszucs:
Nothing to add.

scanners/ios/ios.go
lpusok: That should be happening see `parseResourceSet` function.

_tests/integration/android_test.go
```diff
@@ -157,7 +157,11 @@ var sampleAppsAndroidSDK22SubdirResultYML = fmt.Sprintf(`options:
             env_key: VARIANT
             value_map:
               "":
-                config: android-config
+                title: Application icon
+                value_type: icon
```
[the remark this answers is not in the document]
lpusok: There is already a json field called `value_map` so it is not completely unreasonable. It was meant as a placeholder until final spec is discussed. Open to recommendations.

Gopkg.toml
```diff
@@ -23,12 +27,12 @@
...
 
 [[constraint]]
-  branch = "master"
   name = "github.com/bitrise-io/go-xcode"
+  branch = "master"
 
 [[constraint]]
-  branch = "master"
   name = "github.com/bitrise-io/xcode-project"
+  branch = "feature/asset-catalog"
```
BirmacherAkos: Do not forget to change it back to master after the xcode-project PR merged

scanners/android/icon.go
```diff
@@ -0,0 +1,138 @@
...
+		log.Debugf("Key application not found in manifest file")
+		return nil, nil
+	}
+	ic := app.SelectAttr("android:icon")
+	if ic == nil {
+		log.Debugf("Attribute not found in manifest file")
+		return nil, nil
+	}
+
+	iconPathParts := strings.Split(strings.TrimPrefix(ic.Value, "@"), "/")
```
godrei: NOTE: would be great to have a comment with a usual icon value, to see, why you parse the path this way

scanners/android/icon.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// LookupIcons returns the largest resolution for all potential android icons.
+func LookupIcons(projectDir string, basepath string) (models.Icons, error) {
+	iconPaths, err := lookupIcons(projectDir, basepath)
+	if err != nil {
+		return nil, err
+	}
+
+	icons, err := utility.ConvertPathsToUniqueFileNames(iconPaths, basepath)
```
godrei: NOTE: return utility.ConvertPathsToUniqueFileNames(iconPaths, basepath)

scanners/android/icon.go
godrei: `@mipmap/ic_launcher`

scanners/android/android.go
```diff
@@ -74,15 +86,15 @@ func (scanner *Scanner) Options() (models.OptionNode, models.Warnings, error) {
...
+	return *projectLocationOption, warnings, appIconsAllProjects, nil
 }
 
 // DefaultOptions ...
 func (scanner *Scanner) DefaultOptions() models.OptionNode {
 	projectLocationOption := models.NewOption(ProjectLocationInputTitle, ProjectLocationInputEnvKey)
 	moduleOption := models.NewOption(ModuleInputTitle, ModuleInputEnvKey)
 	variantOption := models.NewOption(VariantInputTitle, VariantInputEnvKey)
-	configOption := models.NewConfigOption(DefaultConfigName)
+	configOption := models.NewConfigOption(DefaultConfigName, []string{})
```
godrei: NOTE: could pass nil instead of an empty array

scanners/android/icon.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	ic := app.SelectAttr("android:icon")
+	if ic == nil {
+		log.Debugf("Attribute not found in manifest file")
+		return nil, nil
+	}
+
+	iconPathParts := strings.Split(strings.TrimPrefix(ic.Value, "@"), "/")
+	if len(iconPathParts) != 2 {
+		return nil, fmt.Errorf("unsupported icon key")
```
godrei: pls include `ic.Value` in the error message to be able fix this issue if occures

scanners/ios/icon.go
```diff
@@ -0,0 +1,121 @@
...
+	target, err := nameToTarget(projectPath, targetName)
+	if err != nil {
+		return models.Icons{}, nil
+	}
+
+	return lookupIconByTarget(projectPath, target, basepath)
+}
+
+func nameToTarget(projectPath string, targetName string) (xcodeproj.Target, error) {
+	project, err := xcodeproj.Open(projectPath)
```
godrei: xcodeproj.Project stores the project path, so pass the project model instead of the path and open it again.

utility/icon.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+	"path/filepath"
+	"sort"
+
+	"github.com/bitrise-io/bitrise-init/models"
+	"github.com/bitrise-io/go-utils/sliceutil"
+)
+
+// ConvertPathsToUniqueFileNames returns a sorted array of unique icons.
+func ConvertPathsToUniqueFileNames(absoluteIconPaths []string, basepath string) (models.Icons, error) {
```
godrei: NOTE: this function does a bit more then i expected based on the function name, what about: `CreateIconDescriptors?`

scanners/ios/resource_set.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+	}
+	return largestIcon, true, nil
+}
+
+func parseResourceSetMetadata(input io.Reader) ([]appIcon, error) {
+	decoder := json.NewDecoder(input)
+	var decoded assetMetadata
+	err := decoder.Decode(&decoded)
+	if err != nil {
```
godrei: NOTE: err could be defined in the if

scanner/result.go
```diff
@@ -12,6 +12,41 @@ import (
 	"github.com/bitrise-io/go-utils/log"
 )
 
+// GenerateScanResult runs the scanner, returns the results and if any platform was detected.
+func GenerateScanResult(searchDir string) (models.ScanResultModel, bool) {
+	scanResult := Config(searchDir)
+
+	platforms := []string{}
```
BirmacherAkos: Note:
We prefer to use `var platforms []string`  instead when it's possible

PR #146, opened 2019-04-17
add project type option to the default fastlane config

godrei:


scanners/fastlane/fastlane.go
```diff
@@ -21,10 +21,14 @@ const (
...
 const (
-	configName        = "fastlane-config"
-	defaultConfigName = "default-fastlane-config"
+	configName              = "fastlane-config"
+	defaultConfigNameFormat = "default-fastlane-%s-config"
 )
 
+const iosPlatform = "ios"
+
+var platforms = []string{iosPlatform, "android"}
```
BirmacherAkos: Why did you use const for iosPlatform only?

lszucs: Is reading `"ios"` from a const but having `"android"` in place intentional?

godrei: because of this: https://github.com/bitrise-io/bitrise-init/pull/146/files#diff-99f7558dc302172389d0f5ed4d4bdbb0R226

godrei: because of this: https://github.com/bitrise-io/bitrise-init/pull/146/files#diff-99f7558dc302172389d0f5ed4d4bdbb0R226

PR #152, opened 2019-05-14
Bugfix/ios icons

lpusok:


cli/cli.go
```diff
@@ -39,24 +34,7 @@ func Run() {
...
-			logLevelStr = "info"
-		}
-
-		level, err := log.ParseLevel(logLevelStr)
-		if err != nil {
-			return err
-		}
-		log.SetLevel(level)
-
+		log.SetEnableDebugLog(true)
```
BirmacherAkos: Do we want to use debug log always?

lpusok: yes

cli/cli.go
```diff
@@ -67,6 +45,7 @@ func Run() {
 	}
 
 	if err := app.Run(os.Args); err != nil {
-		log.Fatal(err)
+		log.TWarnf("%s", err)
```
godrei: why warn? afaik log.Fatal was red colored.

scanners/ios/icon.go
```diff
@@ -23,16 +23,21 @@ func lookupIconBySchemeName(projectPath string, schemeName string, basepath stri
 		return nil, fmt.Errorf("scheme (%s) not found in project", schemeName)
 	}
 
-	mainTarget, err := mainTargetOfScheme(project, scheme.Name)
-
+	mainTarget, found, err := mainTargetOfScheme(project, scheme.Name)
+	if err != nil {
+		return nil, fmt.Errorf("main target not found for project (%s), scheme (%s), error: %s", project.Path, scheme.Name, err)
+	} else if !found {
+		log.TDebugf("No main target found for scheme: %s", scheme.Name)
```
godrei: why did not include the project? above you did

scanners/ios/icon.go
```diff
@@ -85,11 +94,11 @@ func lookupIconByTarget(projectPath string, target xcodeproj.Target, basepath st
 	return icons, nil
 }
 
-func mainTargetOfScheme(proj xcodeproj.XcodeProj, scheme string) (xcodeproj.Target, error) {
+func mainTargetOfScheme(proj xcodeproj.XcodeProj, scheme string) (xcodeproj.Target, bool, error) {
 	projTargets := proj.Proj.Targets
 	sch, ok := proj.Scheme(scheme)
 	if !ok {
-		return xcodeproj.Target{}, fmt.Errorf("Failed to find scheme (%s) in project", scheme)
+		return xcodeproj.Target{}, false, fmt.Errorf("Failed to find scheme (%s) in project", scheme)
```
godrei: i think the only one possible error just codes that the function was not able to find the scheme, what's the point of returning both a bool and an error with this information

PR #160, opened 2019-06-12
Release

lpusok:


Gopkg.toml
```diff
@@ -1,8 +1,4 @@
 
-[[constraint]]
-  name = "github.com/Sirupsen/logrus"
-  version = "1.4.1"
-
 [[constraint]]
   name = "github.com/beevik/etree"
   version = "1.1.0"
```
BirmacherAkos: The bitrise is locked on 1.29.0 here:
https://github.com/bitrise-io/bitrise-init/blob/95f9b5342cc69d411349640001b4d64a9c8d1e3f/Gopkg.toml#L9

But the current version is 1.31.0

lpusok: Updated, did not change to master as if the bitrise.yml format increases, the website could not yet support it.

PR #167, opened 2019-09-04
Adding yarn support for react-native scanner

lpusok:
Would like to support yarn package manager for react-native projects.
Added detection in the common react-native Detect(); the result be used by plain and expo react-native to decide wether npm or yarn install step has to be added.


_tests/integration/reactnative_test.go
```diff
@@ -419,3 +442,184 @@ configs:
...
+          steps:
+          - activate-ssh-key@%s:
+              run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
+          - git-clone@%s: {}
+          - script@%s:
+              title: Do anything with Script step
+          - yarn@%s:
+              inputs:
+              - command: install
+          - npm@%s:
```
BirmacherAkos: I think it should be `yarn test` instead.

BirmacherAkos: And if so do not forget to change the version to yarn here as well:
https://github.com/bitrise-io/bitrise-init/pull/167/files#diff-b2c12688a88be6d0a33b6badc95442dbR463

lpusok:
removed, will not get recreaed

lpusok:
Please merge.

PR #177, opened 2020-01-16
Tool 1345 lock to major step versions

lszucs:
- change default step versions to lock on major version
- fix ci: missing stack

### DIFF
--- a/maintenance/maintenance_test.go
+++ b/maintenance/maintenance_test.go
@@ -24,6 +24,7 @@ func stacks() []string {
 		"osx-xcode-11.0.x",
 		"osx-xcode-11.1.x",
 		"osx-xcode-11.2.x",
+		"osx-xcode-11.3.x",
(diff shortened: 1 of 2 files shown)


maintenance/maintenance_test.go
```diff
@@ -24,6 +24,7 @@ func stacks() []string {
 		"osx-xcode-11.0.x",
 		"osx-xcode-11.1.x",
 		"osx-xcode-11.2.x",
+		"osx-xcode-11.3.x",
```
trapacska: pls mark this update somewhere, like on the card if we have for it as resolved

trapacska: I mean, no. This wold silent our notification to update the stack list in the scanner.

trapacska: please remove this modification from this PR.

PR #180, opened 2020-09-16
Add source = scanner to analytics log data.

godrei:


analytics/analytics.go
```diff
@@ -6,5 +6,9 @@ import (
 
 // LogError sends analytics log using log.RErrorf by setting the stepID.
 func LogError(tag string, data map[string]interface{}, format string, v ...interface{}) {
+	if data == nil {
```
mateherber, line 9: Is nil handled by RErrorf? If not this if would be needed anyway

mateherber (approved):

mateherber (approved):

lpusok (approved):

PR #185, opened 2020-11-11
Add "NoPlatformDetected" recommendation

doomsayer13:


scanner/result.go
```diff
@@ -31,6 +32,7 @@ func GenerateScanResult(searchDir string) (models.ScanResultModel, bool) {
 		scanResult.AddErrorWithRecommendation("general", models.ErrorWithRecommendations{
 			Error: errorMessage,
 			Recommendations: step.Recommendation{
+				"NoPlatformDetected":            true,
```
lpusok, line 35: WDYT of having a for example "error-id" key, which has the value "NoPlatformDetected"? This would make adding additional error IDs easier.

doomsayer13: This was what the others recommended on the meeting today, am I right @godrei & @zoltanszabo-bitrise ?

If we decide to use `error-id` then I think we should change the `BranchRecommendation` structure to this one, and we should use only one type of recommendation, just with different `error-id`?!

zoltanszabo-bitrise: Yes @doomsayer13, this was the proposed solution.

@lpusok my only concern with the `error-id` approach, that it's too generic at this point, not to mention, that the frontend needs to process all recommendations differently. 
By using the `error-id`, it would complicate things further on the FE, since you would have to look into the `error-id` field and make decisions based on the value of it.

Processing the scanner result is already pretty messy on FE, so let's not complicate it even further.

godrei (approved):

scanner/result.go
godrei: i would go with the current solution for now, because of the above mentioned reasons

PR #186, opened 2020-11-13
Ignore Android project scanner errors if we find an other proper project

mateherber:
* Unified gradle file lookup
* Written tests
* Only fail android project scanner if not other project is found proper

### DIFF
--- a/scanners/android/android.go
+++ b/scanners/android/android.go
@@ -36,17 +36,14 @@ func (*Scanner) ExcludedScannerNames() []string {
 func (scanner *Scanner) DetectPlatform(searchDir string) (_ bool, err error) {
 	scanner.SearchDir = searchDir
 
-	scanner.ProjectRoots, err = walkMultipleFiles(searchDir, "build.gradle", "settings.gradle")
+	projectFiles := fileGroups{
+		{"build.gradle", "build.gradle.kts"},
+		{"settings.gradle", "settings.gradle.kts"},
+	}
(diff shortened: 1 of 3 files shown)


Szabadember (approved):

scanners/android/android.go
```diff
@@ -36,18 +36,15 @@ func (*Scanner) ExcludedScannerNames() []string {
...
-	scanner.ProjectRoots, err = walkMultipleFiles(searchDir, "build.gradle", "settings.gradle")
-	if err != nil {
-		return false, fmt.Errorf("failed to search for build.gradle files, error: %s", err)
+	projectFiles := fileGroups{
+		{"build.gradle", "build.gradle.kts"},
+		{"settings.gradle", "settings.gradle.kts"},
 	}
-
-	kotlinRoots, err := walkMultipleFiles(searchDir, "build.gradle.kts", "settings.gradle.kts")
+	scanner.ProjectRoots, err = walkMultipleFileGroups(searchDir, projectFiles)
```
godrei, line 43: we could filter the project roots similarly to this: https://github.com/bitrise-io/bitrise-init/blob/e06545394316a9bfef45f0d774ab957e982ed24e/scanners/ios/xcodeproj.go#L201

so that we can filter project roots inside node_modules for example.

godrei (commented):
Pls see my notes

godrei (approved):

PR #188, opened 2020-11-16
Ignoring Android scanner errors resulting from icon lookup failures.

lpusok:
- If icon format is unsupported, does not fail, instead returns default icons.
- Send failures to analytics Info level log, does not return warnings to user

### DIFF
--- a/analytics/analytics_test.go
+++ b/analytics/analytics_test.go
@@ -1,0 +1,48 @@
+package analytics
+
+import (
+	"reflect"
+	"testing"
+)
+
+func Test_initData(t *testing.T) {
+	type args struct {
+	}
+	tests := []struct {
+		name       string
+		data, want map[string]interface{}
+	}{
+		{
+			name: "Empty data",
+			data: map[string]interface{}{},
+			want: map[string]interface{}{
+				"source": "scanner",
+			},
+		},
+		{
+			name: "nil data",
+			data: nil,
+			want: map[string]interface{}{
+				"source": "scanner",
+			},
+		},
+		{
+			name: "Existing data",
+			data: map[string]interface{}{
+				"A": "B",
+			},
+			want: map[string]interface{}{
+				"source": "scanner",
+				"A":      "B",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			if got := initData(tt.data); !reflect.DeepEqual(got, tt.want) {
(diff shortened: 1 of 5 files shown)


analytics/analytics_test.go
```diff
@@ -0,0 +1,45 @@
...
+			},
+		},
+		{
+			name: "nil data",
+			data: nil,
+			want: map[string]interface{}{
+				"source": "scanner",
+			},
+		},
+		{
```
godrei, line 27: NOTE: might a test case which represents that we set (overwrite the existing value of) `source` would be useful.

scanners/android/icon.go
```diff
@@ -25,30 +26,23 @@ func lookupIconName(manifestPth string) ([]icon, error) {
...
-		log.Debugf("Key application not found in manifest file")
-		return nil, nil
+		return nil, fmt.Errorf("Key application not found in manifest file")
 	}
 	ic := app.SelectAttr("android:icon")
 	if ic == nil {
-		log.Debugf("Attribute not found in manifest file")
-		return nil, nil
+		// Gradle varaibles like ${appIcon} are not supported
+		return nil, fmt.Errorf("Attribute not found in manifest file")
```
godrei, line 45: error string should not be capitalized: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

godrei: NOTE: I think mentioning which attribute (`android:icon`) is missing would make this error more informative

scanners/android/icon.go
```diff
@@ -105,7 +99,7 @@ func lookupIcons(projectDir string, basepath string) ([]string, error) {
 	for _, manifestPath := range manifestPaths {
 		icons, err := lookupIconName(manifestPath)
 		if err != nil {
-			return nil, err
+			analytics.LogInfo("android-icon-lookup", analytics.DetectorErrorData("android", err), "Failed to lookup android icon from manifest")
```
godrei, line 102: missing continue

godrei (changes requested):
Pls see my notes

scanners/android/icon.go
lpusok: append seems to be working without an explicit continue, but maybe the intent is clearer that way?

Szabadember (approved):

scanners/android/icon.go
godrei: i think so

godrei (approved):

PR #193, opened 2020-11-23
Showing a better error message to Ionic Capacitor users

chreno:
### Changes

1. Having cordova.xml is not mandatory anymore for Ionic projects, but if it's missing we show a warning message to the user that atm we only support Ionic Cordova projects.
2. Adding warningsWithRecommendation and errorsWithRecommendation to the not detected projects' output.

Users will see the following while trying to add an Ionic Capacitor project:

image

### DIFF
--- a/scanners/ionic/ionic.go
+++ b/scanners/ionic/ionic.go
@@ -51,7 +51,7 @@ const (
 
 // Scanner ...
 type Scanner struct {
-	cordovaConfigPth    string
+	ionicConfigPath     string
 	relCordovaConfigDir string
 	searchDir           string
 	hasKarmaJasmineTest bool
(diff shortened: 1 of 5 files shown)


scanners/ionic/ionic.go
```diff
@@ -75,56 +75,28 @@ func (scanner *Scanner) DetectPlatform(searchDir string) (bool, error) {
...
+	if ionicConfigPath == "" {
+		ionicConfigPath, err = FilterRootFile(fileList, "ionic.project")
+		if err != nil {
+			return false, fmt.Errorf("failed to check if project is an ionic project, error: %s", err)
+		}
 	}
 
-	if !ionicProjectExist && !ionicConfigExist {
-		log.Printf("no ionic.project file nor ionic.config.json found, seems to be a cordova project")
+	if ionicConfigPath == "" {
```
lpusok, line 92: What is the ionic.project's path (relative), is it is in the root directory?

chreno: It's relative, but includes the file name as well, so it should be fine I guess, shouldn't it?

lpusok: It should be, yes. Usually it is better to use absolute path, where possible.

chreno: Ah ok, thanks! TBH I just copied/modified the existing code here, didn't think about whether it can be a problem, but thanks for highlighting!

lpusok (dismissed):

scanners/ionic/ionic.go
```diff
@@ -222,8 +195,24 @@ func (scanner *Scanner) Options() (models.OptionNode, models.Warnings, models.Ic
...
+		return models.OptionNode{},
+			warnings,
+			nil,
+			fmt.Errorf("failed to search for config.xml file, error: %s", err)
+	}
+
+	log.TPrintf("config.xml: %s", filepath.Join(projectRootDir, "config.xml"))
+
+	if !cordovaConfigExist {
+		warning := fmt.Sprintf("Cordova config.xml not found.")
```
godrei, line 210: Should not this be an error?

We say in this case the user should avoid using the auto configurator, but we allow them to continue. 
Other than that either if the user continues with the generated workflow or goes with manual configuration, the final workflow will contain the same steps (which are not working well with Ionic Capacitator projects).

I think we should do this:
- let the user continue with the auto-generated config (manual config would result in the same steps, just inputs need to be filled by hand)
- update the warning based on the previous and mention that after the add new app flow use workflow editor to configure your project.

godrei (commented):
@chreno pls see my notes

scanners/ionic/ionic.go
chreno: New error message:

`Our auto-configurator only supports Ionic projects with Cordova at the moment. If you’re trying to add a project with Ionic Capacitor, or something else, some Steps in your automatically generated Workflow might fail. To fix this, replace the failing Steps with script Steps in the Workflow editor later.`

godrei (approved):

PR #197, opened 2021-01-06
Improve cocoapods workspace detection

lpusok:
When using react-native, it includes a relative path from node_modules (require_relative '../node_modules/react-native/scripts/react_native_pods').
Fixed the case when node_modules is present. Can not support the case if it is not.

Added 'cocoapods' dependency in the Gemfile, as cocoapods-core is not enough in itself.
Log the reason when can not parse Podfile.


adborbas:
Will open as new PR

PR #198, opened 2021-01-08
Add local.properties check to Android scanner

istvankovacs-bitrise:
https://bitrise.atlassian.net/browse/STEP-164

### Context
Check implemented for `local.properties` file in Android scanner.


scanners/android/utility_test.go
```diff
@@ -171,3 +206,21 @@ func buildMatcher(rootsAndPaths map[string][]string) func(path string) (bool, er
 		return false, nil
 	}
 }
+
+func createProjectDirectory(t *testing.T, containsLocalProperties bool) string {
```
mateherber, line 210: I'm usually not a great fan of unit tests doing side effects (like writing to files, using network or anything) but abstracting the entire FS away for the sake of tests would make this fix much more expensive.

In a distant future we will hopefully use interface implementations for all side effects so we can mock them in tests easily.

mateherber (approved):

PR #199, opened 2021-02-05
Step 170 react cocapods parse error

adborbas:
### Checklist

- Requires _MINOR_ [version update](https://semver.org/)

### Context

Only log "error: failed to get target definition map, error: failed to read target defintion map, error: Pod::DSLError" type of errors instead of returning them as warnings for ReactNative scanner.

Resolves: [STEP-170](https://bitrise.atlassian.net/browse/STEP-170)

### Changes

- Fixed the case when node_modules is present. Can not support the case if it is not.
- Added 'cocoapods' dependency in the Gemfile, as cocoapods-core is not enough in itself.
- Log the reason when can not parse Podfile.

### Investigation details

When using react-native, it includes a relative path from node_modules (require_relative '../node_modules/react-native/scripts/react_native_pods').


scanners/ios/podfile.go
```diff
@@ -250,13 +255,61 @@ func GetWorkspaceProjectMap(podfilePth string, projects []string) (map[string]st
 	}, nil
 }
 
+func (podfileParser podfileParser) podfilelockPath(podfileDir string) (string, error) {
+	podfileLockPth := filepath.Join(podfileDir, "Podfile.lock")
+	if exist, err := pathutil.IsPathExists(podfileLockPth); err != nil {
+		return "", fmt.Errorf("failed to check if Podfile.lock exist, error: %s", err)
```
lpusok, line 261: Please remove ', error', like: `exists: %s", err

scanners/ios/podfile.go
```diff
@@ -81,15 +92,29 @@ end
...
-func getUserDefinedProjectRelavtivePath(podfilePth, cocoapodsVersion string) (string, error) {
-	targetProjectMap, err := getTargetDefinitionProjectMap(podfilePth, cocoapodsVersion)
+func (podfileParser podfileParser) shouldRaiseReadDefinitionError(err string) bool {
+	if err == "" {
+		return false
+	}
+
+	isInvalidPodfileError := stringutil.CaseInsensitiveContains(err, "Pod::DSLError")
+	if isInvalidPodfileError && podfileParser.suppressPodFileParseError {
+		log.Warnf("could not parse podfile, defult behaviour will be continued, %s", err)
```
lpusok, line 109: Log functions with T prefix are used in scanner in most places. Would move warning second part in a new line for better readebility.

log.TWarnf("Could not parse podfile: %s", err)
log.TWarnf("Will continue using defualt Cocapods paths.")

scanners/ios/podfile.go
```diff
@@ -250,13 +255,61 @@ func GetWorkspaceProjectMap(podfilePth string, projects []string) (map[string]st
...
+}
+
 // MergePodWorkspaceProjectMap ...
 // Previously we separated standalone projects and workspaces.
 // But pod workspace-project map may define workspace which is not in the repository, but will be created by `pod install`.
 // Related project should be found in the standalone projects list.
 // We will create this workspace model, join the related project and remove this project from standlone projects.
 // If workspace is in the repository, both workspace and project should be find in the input lists.
-func MergePodWorkspaceProjectMap(podWorkspaceProjectMap map[string]string, standaloneProjects []xcodeproj.ProjectModel, workspaces []xcodeproj.WorkspaceModel) ([]xcodeproj.ProjectModel, []xcodeproj.WorkspaceModel, error) {
+func (podfileParser podfileParser) MergePodWorkspaceProjectMap(podWorkspaceProjectMap map[string]string, standaloneProjects []xcodeproj.ProjectModel, workspaces []xcodeproj.WorkspaceModel) ([]xcodeproj.ProjectModel, []xcodeproj.WorkspaceModel, error) {
```
lpusok, line 312: Why does this belong to podfileParser? As far as I can see it deos not use any proprty from it.

scanners/ios/podfile.go
```diff
@@ -250,13 +255,61 @@ func GetWorkspaceProjectMap(podfilePth string, projects []string) (map[string]st
...
+
+func (podfileParser podfileParser) cocoapodsVersion(podfileLockPth string) (string, error) {
+	cocoapodsVersion := ""
+
+	if podfileLockPth != "" {
+		version, err := GemVersionFromGemfileLock("cocoapods", podfileLockPth)
+		if err != nil {
+			return "", fmt.Errorf("failed to read cocoapods version from %s, error: %s", podfileLockPth, err)
+		}
+		cocoapodsVersion = version
```
lpusok, line 282: return version, err

scanners/ios/podfile.go
```diff
@@ -250,13 +255,61 @@ func GetWorkspaceProjectMap(podfilePth string, projects []string) (map[string]st
...
+	}
+
+	return cocoapodsVersion, nil
```
lpusok, line 285: return "", nil
This is simpler than having to define cocapodsVersion

scanners/ios/podfile.go
```diff
@@ -81,15 +92,29 @@ end
...
 }
 
-func getUserDefinedProjectRelavtivePath(podfilePth, cocoapodsVersion string) (string, error) {
-	targetProjectMap, err := getTargetDefinitionProjectMap(podfilePth, cocoapodsVersion)
+func (podfileParser podfileParser) shouldRaiseReadDefinitionError(err string) bool {
+	if err == "" {
+		return false
+	}
+
+	isInvalidPodfileError := stringutil.CaseInsensitiveContains(err, "Pod::DSLError")
```
lpusok, line 107: Why is the case insensitive match needed?

lpusok (changes requested):

scanners/ios/podfile.go
adborbas: Btw what is the difference between `TWarnf` and `Warnf`?

scanners/ios/podfile.go
adborbas: Nice catch 👍

scanners/ios/podfile.go
adborbas: You are right, it is no longer needed. I was comparing different strings for which it made sense.

scanners/ios/podfile.go
```diff
@@ -104,9 +103,10 @@ func (podfileParser podfileParser) shouldRaiseReadDefinitionError(err string) bo
 		return false
 	}
 
-	isInvalidPodfileError := stringutil.CaseInsensitiveContains(err, "Pod::DSLError")
+	isInvalidPodfileError := strings.Contains(err, "Pod::DSLError")
 	if isInvalidPodfileError && podfileParser.suppressPodFileParseError {
-		log.Warnf("could not parse podfile, defult behaviour will be continued, %s", err)
+		log.TWarnf("Could not parse podfile: %s", err)
+		log.TWarnf("Will continue using defualt Cocoapods paths.")
```
lpusok, line 109: typo in defualt -> default

lpusok (approved):

PR #209, opened 2021-07-14
Adds log statements to the android project scanner

benbitrise:
image

I was troubleshooting the project scanner, and it was unclear to me if the android project scanner found any files. This PR adds some log statements to make this easier to quickly get info.


maintenance/maintenance_test.go
```diff
@@ -28,6 +28,7 @@ func stacks() []string {
 		"osx-xcode-12.3.x",
 		"osx-xcode-12.4.x",
 		"osx-xcode-12.5.x",
+		"osx-xcode-13.0.x",
```
benbitrise, line 31: Fixes a failing unit test by adding this live stack

PR #213, opened 2021-09-13
Deploy to Bitrise.io 2.0.0

Bence1001:
## Changes
**Update Deploy to Bitrise.io step major version to 2.**


ofalvai (dismissed):

maintenance/maintenance_test.go
```diff
@@ -1,83 +0,0 @@
-package maintenance
```
Bence1001, line 1: I removed this test because it seems to be unnecessary.

ofalvai (approved):

PR #214, opened 2021-09-14
Do not specify module in Android Lint and Android Unit Test

Bence1001:
## Context
* When a new Android app is added to Bitrise, we create default workflows for the user: `primary` and `deploy`.
* In these workflows we use the Android Lint and Android Unit Test steps, where the `Module` input is set to `$MODULE`. This means tests and lint checks will only be performed against the module that was specified by the user when the project's Bitrise config was generated (e.g. when the app was added to Bitrise).
* In multi-module Android projects, unit tests and lint issues can be in multiple modules, not just in the one that was initially provided by the user.
* Therefore, it doesn't make much sense to run lint checks and unit tests only in that module by default.

## Changes
* Do not specify any modules in the generated workflows' Android Lint and Android Unit Test steps.

### DIFF
--- a/scanners/android/utility.go
+++ b/scanners/android/utility.go
@@ -16,7 +16,6 @@ type fileGroups [][]string
 var pathUtilIsPathExists = pathutil.IsPathExists
 var filePathWalk = filepath.Walk
 
-// Constants ...
 const (
 	ScannerName       = "android"
 	ConfigName        = "android-config"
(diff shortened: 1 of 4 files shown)


scanners/android/utility.go
```diff
@@ -39,9 +38,7 @@ const (
 	ModuleInputTitle   = "Module"
 	ModuleInputSummary = "Modules provide a container for your Android project's source code, resource files, and app level settings, such as the module-level build file and Android manifest file. Each module can be independently built, tested, and debugged. You can add new modules to your Bitrise builds at any time."
 
-	GradlewPathInputKey    = "gradlew_path"
-	GradlewPathInputEnvKey = "GRADLEW_PATH"
-	GradlewPathInputTitle  = "Gradlew file path"
```
Bence1001, line 44: These constants were unused.

ofalvai (approved):

mateherber (dismissed):

mateherber (approved):

ofalvai (approved):

PR #217, opened 2021-11-01
[MOB-222] Remove the Xamarin scanner

tothszabi:
This PR removes the Xamarin scanner and all the constants and any Xamarin reference from the codebase. The major version is bumped because this is a breaking change.


ofalvai (approved):
LGTM, but I would also wait for another review from someone who is more familiar with this codebase than me.

I also checked `xamarin` references in this repo [with Sourcegraph](https://sourcegraph.com/search?q=context:global+repo:bitrise.io/bitrise-init+xamarin+rev:refs/heads/MOB-222-xamarin-removal&patternType=literal) and I think we are good with this PR!

mateherber (approved):
Do the website (add-new-app-flow) and project type selector need PRs as well?
Will we deprecate the `go-xamarin` lib too?

ofalvai (approved):

tothszabi:
@mateherber they will need PRs but will be handled by other teams and only at the end of January when the actual removal will happen. 

I will add a deprecation notice to `go-xamarin` as well because it is only used by our soon to be removed xamarin related steps.

PR #219, opened 2022-01-24
Update Android deploy WF

lpusok:
### Checklist

- [X] I've read and followed the [Contribution Guidelines](https://github.com/bitrise-steplib/.github/blob/main/CONTRIBUTING.md)
- [X] `step.yml` and `README.md` is updated with the changes (if needed)

### Version

Requires a *PATCH* [version update](https://semver.org/)

### Context

Updated Android deploy WF.

Resolves: https://bitrise.atlassian.net/browse/STEP-1747

### Changes

- Introduces the isPrivateRepo paramter in the Config method of ScannerInterface.
- Fixed an unrelated Cocoapods test.

### DIFF
--- a/scanners/macos/macos.go
+++ b/scanners/macos/macos.go
@@ -60,7 +60,7 @@ func (Scanner) DefaultOptions() models.OptionNode {
 }
 
 // Configs ...
-func (scanner *Scanner) Configs() (models.BitriseConfigMap, error) {
+func (scanner *Scanner) Configs(_ bool) (models.BitriseConfigMap, error) {
 	return ios.GenerateConfig(ios.XcodeProjectTypeMacOS, scanner.configDescriptors, true)
 }
 
--- a/steps/steps.go
+++ b/steps/steps.go
@@ -7,6 +7,12 @@ import (
 	stepmanModels "github.com/bitrise-io/stepman/models"
 )
 
+// PrepareListParams describes the default prepare Step options.
+type PrepareListParams struct {
+	ShouldIncludeCache       bool
+	ShouldIncludeActivateSSH bool
+}
+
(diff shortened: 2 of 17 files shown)


scanners/macos/macos.go
```diff
@@ -60,7 +60,7 @@ func (Scanner) DefaultOptions() models.OptionNode {
 }
 
 // Configs ...
-func (scanner *Scanner) Configs() (models.BitriseConfigMap, error) {
+func (scanner *Scanner) Configs(_ bool) (models.BitriseConfigMap, error) {
```
tothszabi, line 63: TIL that you can use the blank identifier in the function signature as well. Nice.

steps/steps.go
```diff
@@ -45,6 +51,23 @@ func DefaultPrepareStepList(isIncludeCache bool) []bitriseModels.StepListItemMod
 	return append(stepList, ScriptSteplistItem(ScriptDefaultTitle))
 }
 
+// DefaultPrepareStepListV2 ...
+func DefaultPrepareStepListV2(params PrepareListParams) []bitriseModels.StepListItemModel {
```
tothszabi, line 55: This comment applies to the `DefaultDeployStepListV2` function too. What if we modify the original functions instead and let all of the workflows benefit from them?

Putting the deploy step last will decrease the failure rate for all of them. And it is true that SSH key activation skipping will require a little extra work from the scanners and immediate benefit. 

So I think I would change the deploy one for all of the workflow and the other is up for debate. What do you think?

lpusok: I would prefer to apply the change in smaller increments, as it also requires manually editing the e2e tests.

tothszabi (approved):

PR #221, opened 2022-01-26
[STEP-1750 & STEP-1749] Update iOS default workflows

tothszabi:
Tickets:
- https://bitrise.atlassian.net/browse/STEP-1750
- https://bitrise.atlassian.net/browse/STEP-1749

Changes:
- Appclip related logic is extracted to a dedicated file
- The workflow generation is also extracted into a separate file
    - The three different hard coded paths are unified
    - The common elements are shared between the two workflows
- Xcode build for test step is added 
- The integration tests are updated to reflect the changes

### DIFF
--- a/_tests/integration/fastlane_test.go
+++ b/_tests/integration/fastlane_test.go
@@ -175,9 +175,13 @@ configs:
           - cache-pull@%s: {}
           - xcode-test@%s:
               inputs:
+              - project_path: $BITRISE_PROJECT_PATH
+              - scheme: $BITRISE_SCHEME
               - test_repetition_mode: retry_on_failure
           - xcode-archive@%s:
               inputs:
+              - project_path: $BITRISE_PROJECT_PATH
+              - scheme: $BITRISE_SCHEME
               - distribution_method: $BITRISE_DISTRIBUTION_METHOD
               - automatic_code_signing: api-key
           - cache-push@%s: {}
@@ -194,6 +198,8 @@ configs:
           - cache-pull@%s: {}
           - xcode-test@%s:
               inputs:
+              - project_path: $BITRISE_PROJECT_PATH
+              - scheme: $BITRISE_SCHEME
               - test_repetition_mode: retry_on_failure
           - cache-push@%s: {}
           - deploy-to-bitrise-io@%s: {}
--- a/_tests/integration/fastlane_test.go
+++ b/_tests/integration/fastlane_test.go
@@ -178,6 +178,7 @@ configs:
               - test_repetition_mode: retry_on_failure
           - xcode-archive@%s:
               inputs:
+              - distribution_method: $BITRISE_DISTRIBUTION_METHOD
               - automatic_code_signing: api-key
           - cache-push@%s: {}
           - deploy-to-bitrise-io@%s: {}
--- a/_tests/integration/fastlane_test.go
+++ b/_tests/integration/fastlane_test.go
@@ -51,21 +51,17 @@ var fastlaneVersions = []interface{}{
 	steps.ActivateSSHKeyVersion,
 	steps.GitCloneVersion,
 	steps.CachePullVersion,
-	steps.ScriptVersion,
-	steps.CertificateAndProfileInstallerVersion,
 	steps.XcodeTestVersion,
 	steps.XcodeArchiveVersion,
-	steps.DeployToBitriseIoVersion,
 	steps.CachePushVersion,
+	steps.DeployToBitriseIoVersion,
 
 	steps.ActivateSSHKeyVersion,
 	steps.GitCloneVersion,
 	steps.CachePullVersion,
-	steps.ScriptVersion,
-	steps.CertificateAndProfileInstallerVersion,
 	steps.XcodeTestVersion,
-	steps.DeployToBitriseIoVersion,
 	steps.CachePushVersion,
+	steps.DeployToBitriseIoVersion,
 }
 
 var fastlaneResultYML = fmt.Sprintf(`options:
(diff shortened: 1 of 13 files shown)


_tests/integration/ios_test.go
```diff
@@ -694,26 +700,43 @@ configs:
...
           - git-clone@%s: {}
           - cache-pull@%s: {}
-          - script@%s:
-              title: Do anything with Script step
-          - certificate-and-profile-installer@%s: {}
           - xcode-archive@%s:
               inputs:
-              - project_path: $BITRISE_PROJECT_PATH
-              - scheme: $BITRISE_SCHEME
-              - distribution_method: $BITRISE_DISTRIBUTION_METHOD
```
lpusok, line 710: Discussed that *distribution_method*  has `development` not `$BITRISE_DISTRIBUTION_METHOD` as default value, needs to be included.

tothszabi: Thanks for pointing it out. Fixed it.

_tests/integration/fastlane_test.go
```diff
@@ -165,40 +161,48 @@ configs:
...
               - project_path: $BITRISE_PROJECT_PATH
               - scheme: $BITRISE_SCHEME
               - distribution_method: $BITRISE_DISTRIBUTION_METHOD
-          - deploy-to-bitrise-io@%s: {}
+              - automatic_code_signing: api-key
           - cache-push@%s: {}
+          - deploy-to-bitrise-io@%s: {}
         primary:
+          description: |
+            The workflow executes the tests. And if there are none then only builds the project.
```
godrei, line 190: NOTE: we know whether we found tests or not. Should not we provide a specific description?

scanners/ios/appclip.go
```diff
@@ -0,0 +1,36 @@
+package ios
+
+import (
+	"github.com/bitrise-io/bitrise-init/models"
+	"github.com/bitrise-io/bitrise-init/steps"
+	envmanModels "github.com/bitrise-io/envman/models"
+	"github.com/bitrise-io/go-xcode/xcodeproj"
+)
+
+func schemeHasAppClipTarget(scheme xcodeproj.SchemeModel, targets []xcodeproj.TargetModel) bool {
```
godrei, line 10: NOTE: we have an issue with detecting app-clips, not sure if this code has the same bug, but worth checking: https://bitrise.atlassian.net/browse/STEP-1754

_tests/integration/fastlane_test.go
```diff
@@ -165,40 +161,48 @@ configs:
         workflow: primary
       workflows:
         deploy:
+          description: |
+            Deploys app using *Deploy to bitrise.io* step.
```
godrei, line 165: NOTE: might i would talk a bit more about this wf: tests, build, and deploys; the test has `retry_on_failure` test repetition mode enabled... 
might be worth passing the description to tech writers too.

_tests/integration/mac_test.go
```diff
@@ -113,9 +109,6 @@ configs:
               run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
           - git-clone@%s: {}
           - cache-pull@%s: {}
-          - script@%s:
```
godrei, line 116: Should not we add description to macOS wfs too? It is almost the same as the ios description.

scanners/ios/utility.go
```diff
@@ -615,101 +613,19 @@ func GenerateDefaultOptions(projectType XcodeProjectType) models.OptionNode {
...
-				appendExportAppClipStep(configBuilder, models.DeployWorkflowID)
-			}
-		case XcodeProjectTypeMacOS:
-			configBuilder.AppendStepListItemsTo(models.DeployWorkflowID, steps.XcodeTestMacStepListItem(xcodeStepInputModels...))
-			configBuilder.AppendStepListItemsTo(models.DeployWorkflowID, steps.XcodeArchiveMacStepListItem(xcodeArchiveMacStepInputModels...))
-		}
-
-		configBuilder.AppendStepListItemsTo(models.DeployWorkflowID, steps.DefaultDeployStepList(isIncludeCache)...)
-	}
+	createPrimaryWorkflow(projectType, configBuilder, isPrivateRepository, includeCache, missingSharedSchemes, hasTest, hasPodfile, carthageCommand)
```
godrei, line 627: NOTE: it might be better to use a single struct param, instead of this much direct params

scanners/ios/appclip.go
lpusok: It looks to me that the code was moved here from utility.go without changes (except *appendExportAppClipStep*), would keep the bugfix in a followup PR to limit the scope.

_tests/integration/mac_test.go
```diff
@@ -113,9 +109,6 @@ configs:
...
-              title: Do anything with Script step
-          - certificate-and-profile-installer@%s: {}
```
lpusok, line 118: Mac Steps still need certificate and profile installer as there is no other Option available.

lpusok (changes requested):

_tests/integration/mac_test.go
tothszabi: Thanks for catching this.

_tests/integration/mac_test.go
tothszabi: We do not have a macOS section in our getting started guides so even if the workflows are similar we could not link to anything useful. This is why I decided to not to change the macOS default workflow description.

lpusok (dismissed):

godrei (approved):

PR #222, opened 2022-02-01
Updating react-native workflows

lpusok:
### Checklist
- [x] I've read and followed the [Contribution Guidelines](https://github.com/bitrise-steplib/.github/blob/main/CONTRIBUTING.md)
- [-] `step.yml` and `README.md` is updated with the changes (if needed)

### Version

### Context
Updates react-native workflows:
- Includes _npm_/_yarn_ `test` in **deploy** workflow
- **primary** and **deploy** workflow always created. If tests are not found in the project **primary** workflow contains _npm_/_yarn_ `install` only, but not building the native apps. Previously **primary** took **deploy**'s place in these cases.
- Empty Script Step removed
- _Certificate and profile installer_ Step removed, _Xcode Archive & Export for iOS_'s Automatic code signing option is enabled instead.
- Fixes issue with an invalid workflow generated for projects without tests. (The same logic as for tests is used.)

Resolves
https://bitrise.atlassian.net/browse/STEP-1744
https://bitrise.atlassian.net/browse/STEP-1742

### Changes

Not in scope yet:
- Default workflow updates
- Description updates
- Testing with other sample apps
- Expo primary workflow update

### DIFF
--- a/_tests/integration/reactnative_test.go
+++ b/_tests/integration/reactnative_test.go
@@ -265,8 +265,7 @@ configs:
             next change in your repository that matches any of your trigger map event will
             start **deploy** workflow.\n"
           steps:
-          - activate-ssh-key@%s:
-              run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
+          - activate-ssh-key@%s: {}
           - git-clone@%s: {}
           - npm@%s:
               inputs:
@@ -292,8 +291,7 @@ configs:
           - deploy-to-bitrise-io@%s: {}
         primary:
           steps:
-          - activate-ssh-key@%s:
-              run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
+          - activate-ssh-key@%s: {}
           - git-clone@%s: {}
           - npm@%s:
               inputs:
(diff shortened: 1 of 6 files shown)


_tests/integration/reactnative_test.go
```diff
@@ -447,22 +490,192 @@ warnings_with_recommendations:
...
+              - project_path: $BITRISE_PROJECT_PATH
+              - scheme: $BITRISE_SCHEME
+              - distribution_method: $BITRISE_DISTRIBUTION_METHOD
+              - configuration: Release
+              - automatic_code_signing: api-key
+          - deploy-to-bitrise-io@%s: {}
+        primary:
+          steps:
+          - activate-ssh-key@%s:
+              run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
```
ofalvai, line 653: If you pull the latest master, this will be removed from the code generating the YAML, you should adjust these tests too

scanners/reactnative/plain.go
```diff
@@ -404,3 +317,22 @@ func (scanner *Scanner) defaultConfigs() (models.BitriseConfigMap, error) {
 
 	return configMap, nil
 }
+
+func (scanner *Scanner) addTestSteps(configBuilder *models.ConfigBuilderModel, workflowID models.WorkflowID, workDir string) {
```
ofalvai, line 321: I think it would be better if this function simply returned the list of steps instead of directly calling the config builder through a pointer. Like `steps.DefaultDeployStepListV2()`

ofalvai (changes requested):
LGTM, but had a few minor comments

PR #223, opened 2022-02-01
Update Flutter primary workflow

ofalvai:
### Context

Update the primary Flutter workflow creation by removing some steps and simplifying detection.

#### High-level overview

- Remove Script step in all possible cases
- Remove Flutter Analyze step in all possible cases
- Add primary workflow description
- Don't ask the user about the presence of tests. This information is detected by the scanner and we can insert/omit the Flutter Test step just like before.
- Don't insert SSH key step if the Git repo is public (as it's not needed for public repos)
- Swap Cache Push and Deploy to Bitrise.io steps

#### Details

See my line-by-line comments in the diff view

### DIFF
--- a/_tests/integration/flutter_test.go
+++ b/_tests/integration/flutter_test.go
@@ -289,8 +289,7 @@ configs:
             Next steps:
             - Check out [Getting started with Flutter apps](https://devcenter.bitrise.io/en/getting-started/getting-started-with-flutter-apps.html).
           steps:
-          - activate-ssh-key@%s:
-              run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
+          - activate-ssh-key@%s: {}
           - git-clone@%s: {}
           - flutter-installer@%s:
               inputs:
@@ -342,8 +341,7 @@ configs:
             Next steps:
             - Check out [Getting started with Flutter apps](https://devcenter.bitrise.io/en/getting-started/getting-started-with-flutter-apps.html).
           steps:
-          - activate-ssh-key@%s:
-              run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
+          - activate-ssh-key@%s: {}
           - git-clone@%s: {}
           - flutter-installer@%s:
               inputs:
@@ -395,8 +393,7 @@ configs:
             Next steps:
             - Check out [Getting started with Flutter apps](https://devcenter.bitrise.io/en/getting-started/getting-started-with-flutter-apps.html).
           steps:
-          - activate-ssh-key@%s:
-              run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
+          - activate-ssh-key@%s: {}
           - git-clone@%s: {}
           - flutter-installer@%s:
               inputs:
@@ -444,8 +441,7 @@ configs:
             Next steps:
             - Check out [Getting started with Flutter apps](https://devcenter.bitrise.io/en/getting-started/getting-started-with-flutter-apps.html).
           steps:
-          - activate-ssh-key@%s:
-              run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
+          - activate-ssh-key@%s: {}
           - git-clone@%s: {}
           - flutter-installer@%s:
               inputs:
@@ -503,8 +499,7 @@ configs:
             Next steps:
             - Check out [Getting started with Flutter apps](https://devcenter.bitrise.io/en/getting-started/getting-started-with-flutter-apps.html).
           steps:
-          - activate-ssh-key@%s:
-              run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
+          - activate-ssh-key@%s: {}
           - git-clone@%s: {}
           - flutter-installer@%s:
               inputs:
@@ -562,8 +557,7 @@ configs:
             Next steps:
             - Check out [Getting started with Flutter apps](https://devcenter.bitrise.io/en/getting-started/getting-started-with-flutter-apps.html).
           steps:
-          - activate-ssh-key@%s:
-              run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
+          - activate-ssh-key@%s: {}
           - git-clone@%s: {}
           - flutter-installer@%s:
               inputs:
@@ -760,8 +754,7 @@ configs:
             Next steps:
             - Check out [Getting started with Flutter apps](https://devcenter.bitrise.io/en/getting-started/getting-started-with-flutter-apps.html).
           steps:
-          - activate-ssh-key@%s:
-              run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
+          - activate-ssh-key@%s: {}
           - git-clone@%s: {}
           - flutter-installer@%s:
               inputs:
@@ -813,8 +806,7 @@ configs:
             Next steps:
             - Check out [Getting started with Flutter apps](https://devcenter.bitrise.io/en/getting-started/getting-started-with-flutter-apps.html).
           steps:
-          - activate-ssh-key@%s:
-              run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
+          - activate-ssh-key@%s: {}
           - git-clone@%s: {}
           - flutter-installer@%s:
               inputs:
@@ -866,8 +858,7 @@ configs:
             Next steps:
             - Check out [Getting started with Flutter apps](https://devcenter.bitrise.io/en/getting-started/getting-started-with-flutter-apps.html).
           steps:
-          - activate-ssh-key@%s:
-              run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
+          - activate-ssh-key@%s: {}
           - git-clone@%s: {}
           - flutter-installer@%s:
               inputs:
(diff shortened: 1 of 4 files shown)


scanners/flutter/flutter.go
```diff
@@ -249,74 +244,35 @@ func (scanner *Scanner) Options() (models.OptionNode, models.Warnings, models.Ic
 	flutterProjectLocationOption := models.NewOption(projectLocationInputTitle, projectLocationInputSummary, projectLocationInputEnvKey, models.TypeSelector)
 
 	for _, project := range scanner.projects {
+		var testKey string
 		if project.hasTest {
-			flutterProjectHasTestOption := models.NewOption(testsInputTitle, testsInputSummary, "", models.TypeSelector)
```
ofalvai, line 253: Remove the test question completely

scanners/flutter/flutter.go
```diff
@@ -249,74 +244,35 @@ func (scanner *Scanner) Options() (models.OptionNode, models.Warnings, models.Ic
...
-					cfg += "-test"
-				}
+			testKey = "test"
+		} else {
+			testKey = "notest"
+		}
+		cfg := configName + "-" + testKey
 
-				if project.hasIosProject || project.hasAndroidProject {
-					if project.hasIosProject {
```
ofalvai, line 263: This nested if was unnecessary complex. The changed lines below are caused by the indent change

scanners/flutter/flutter.go
```diff
@@ -249,74 +244,35 @@ func (scanner *Scanner) Options() (models.OptionNode, models.Warnings, models.Ic
...
-						projectPathOption := models.NewOption(ios.ProjectPathInputTitle, ios.ProjectPathInputSummary, ios.ProjectPathInputEnvKey, models.TypeSelector)
-						flutterProjectHasTestOption.AddOption(v, projectPathOption)
+		if project.hasIosProject {
+			xcodeProjectPathOption := models.NewOption(ios.ProjectPathInputTitle, ios.ProjectPathInputSummary, ios.ProjectPathInputEnvKey, models.TypeSelector)
```
ofalvai, line 256: Renamed `projectPathOption` to make it obvious that it's iOS specific

scanners/flutter/flutter.go
```diff
@@ -249,74 +244,35 @@ func (scanner *Scanner) Options() (models.OptionNode, models.Warnings, models.Ic
...
-								}
-							}
-						}
-					} else {
+					for _, exportMethod := range ios.IosExportMethods {
 						configOption := models.NewConfigOption(cfg+"-app-"+getBuildablePlatform(project.hasAndroidProject, project.hasIosProject), nil)
-						flutterProjectHasTestOption.AddOption(v, configOption)
+						distributionMethodOption.AddConfig(exportMethod, configOption)
 					}
-				} else {
```
ofalvai, line 285: This path is reached when the repo doesn't have either iOS or Android projects. The chance for this is pretty small, so we don't care about this case, and `getBuildablePlatform()` will identify this as `both`. It means that the deploy workflow will contain both iOS and Android steps.

scanners/flutter/flutter.go
```diff
@@ -249,74 +244,35 @@ func (scanner *Scanner) Options() (models.OptionNode, models.Warnings, models.Ic
...
-					configOption := models.NewConfigOption(cfg, nil)
-					flutterProjectHasTestOption.AddOption(v, configOption)
 				}
 			}
 		} else {
-			cfg := configName
```
ofalvai, line 291: This is the `!project.hasTest` path, it was exactly the same as the above code block.

scanners/flutter/flutter.go
```diff
@@ -338,76 +294,70 @@ func getBuildablePlatform(hasAndroidProject, hasIosProject bool) string {
 func (Scanner) DefaultOptions() models.OptionNode {
 	flutterProjectLocationOption := models.NewOption(projectLocationInputTitle, projectLocationInputSummary, projectLocationInputEnvKey, models.TypeUserInput)
 
-	flutterProjectHasTestOption := models.NewOption(testsInputTitle, testsInputSummary, "", models.TypeSelector)
-	flutterProjectLocationOption.AddOption("", flutterProjectHasTestOption)
+	cfg := configName + "-test"
```
ofalvai, line 297: The default is to assume tests

scanners/flutter/flutter.go
```diff
@@ -27,19 +27,14 @@ const (
...
 	platformInputTitle          = "Platform"
 	projectLocationInputSummary = "The path to your Flutter project, stored as an Environment Variable. In your Workflows, you can specify paths relative to this path. You can change this at any time."
-	testsInputSummary           = "Our Flutter Test Step can run the tests found in your project's repository."
 	platformInputSummary        = "The target platform for your first build. Your options are iOS, Android, both, or neither. You can change this in your Env Vars at any time."
 	installerUpdateFlutterKey   = "is_update"
 )
 
 var (
 	platforms = []string{
-		"none",
```
ofalvai, line 42: This was not possible anyway, `getBuildablePlatforms()` never returns this value.

scanners/flutter/flutter.go
```diff
@@ -338,76 +294,70 @@ func getBuildablePlatform(hasAndroidProject, hasIosProject bool) string {
...
 
-					for _, exportMethod := range ios.IosExportMethods {
-						configOption := models.NewConfigOption(cfg+"-app-"+platform, nil)
-						distributionMethodOption.AddConfig(exportMethod, configOption)
-					}
-				} else {
-					configOption := models.NewConfigOption(cfg+"-app-"+platform, nil)
-					flutterPlatformOption.AddConfig(platform, configOption)
-				}
-			} else {
```
ofalvai, line 372: `none` platform path, it was not reachable anyway.

scanners/flutter/flutter.go
```diff
@@ -338,76 +294,70 @@ func getBuildablePlatform(hasAndroidProject, hasIosProject bool) string {
...
+	return generateConfigMap(false)
+}
+
+func generateConfigMap(isPrivateRepository bool) (models.BitriseConfigMap, error) {
 	configs := models.BitriseConfigMap{}
 
 	for _, variant := range []struct {
 		configID string
 		test     bool
-		deploy   bool
```
ofalvai, line 394: This is not important anymore, we'll generate a deploy workflow in all cases, even if the project has no Android nor iOS platforms.

scanners/flutter/flutter.go
```diff
@@ -338,76 +294,70 @@ func getBuildablePlatform(hasAndroidProject, hasIosProject bool) string {
...
 		platform string
 	}{
-		{test: false, deploy: false, configID: configName},
-		{test: true, deploy: false, configID: configName + "-test"},
```
ofalvai, line 398: Removing the two combinations where `deploy` was false

scanners/flutter/flutter.go
```diff
@@ -416,60 +366,53 @@ func (scanner Scanner) DefaultConfigs() (models.BitriseConfigMap, error) {
...
 			))
 		}
 
-		configBuilder.AppendStepListItemsTo(models.PrimaryWorkflowID, steps.DefaultDeployStepList(true)...)
+		configBuilder.AppendStepListItemsTo(models.PrimaryWorkflowID, steps.DefaultDeployStepListV2(true)...)
 
 		// deploy
+		configBuilder.AppendStepListItemsTo(models.DeployWorkflowID, steps.DefaultPrepareStepList(false)...)
 
-		if variant.deploy {
```
ofalvai, line 433: This messes up the diff of the lines below, but essentially we move the code block out of the `if variant.deploy`

_tests/integration/flutter_test.go
```diff
@@ -74,18 +78,7 @@ func TestFlutter(t *testing.T) {
 }
 
 var flutterSampleAppVersions = []interface{}{
-	// flutter-config
```
ofalvai, line 77: This is one of the two option paths that we remove

_tests/integration/flutter_test.go
```diff
@@ -248,109 +217,42 @@ var flutterSampleAppResultYML = fmt.Sprintf(`options:
...
-        title: Run tests found in the project
-        summary: Our Flutter Test Step can run the tests found in your project's repository.
+        title: Project or Workspace path
+        summary: The location of your Xcode project or Xcode workspace files, stored
+          as an Environment Variable. In your Workflows, you can specify paths relative
+          to this path.
+        env_key: BITRISE_PROJECT_PATH
         type: selector
         value_map:
-          "no":
```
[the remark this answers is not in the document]
lpusok: Not exactly the same, one had `config: flutter-config-app-both` and the other `config: flutter-config-test-app-both`. (test is extra)

scanners/flutter/flutter.go
```diff
@@ -27,19 +27,14 @@ const (
 	defaultIOSConfiguration     = "Release"
 	projectLocationInputEnvKey  = "BITRISE_FLUTTER_PROJECT_LOCATION"
 	projectLocationInputTitle   = "Project location"
-	projectTypeInputEnvKey      = "BITRISE_FLUTTER_PROJECT_TYPE"
-	projectTypeInputTitle       = "Project Type"
-	testsInputTitle             = "Run tests found in the project"
 	platformInputTitle          = "Platform"
 	projectLocationInputSummary = "The path to your Flutter project, stored as an Environment Variable. In your Workflows, you can specify paths relative to this path. You can change this at any time."
-	testsInputSummary           = "Our Flutter Test Step can run the tests found in your project's repository."
```
lpusok, line 35: What happens for the default workflow, are tests always included in that case?

tothszabi (approved):

PR #225, opened 2022-02-03
Update Flutter deploy workflow

ofalvai:
### Context

Updating and simplifying the Flutter workflows. Follow-up to #223 

### Changes

- Remove Xcode archive step from the workflow as the Flutter build command has this functionality now for a while
- Set `ios_build_type` to `archive` in the Flutter build step if platform is `ios` or `both` to build an xcarchive
- Remove empty Script step
- Remove scanner options related to the Xcode archive step, such as the distribution method. These options are no longer needed, because the Xcode Archive step is removed, plus the deploy workflow needs some manual configuration anyway before the first successful build
- Add workflow description

### DIFF
--- a/scanners/flutter/flutter.go
+++ b/scanners/flutter/flutter.go
@@ -24,7 +24,6 @@ const (
 	configName                  = "flutter-config"
 	projectLocationInputKey     = "project_location"
 	platformInputKey            = "platform"
-	defaultIOSConfiguration     = "Release"
 	projectLocationInputEnvKey  = "BITRISE_FLUTTER_PROJECT_LOCATION"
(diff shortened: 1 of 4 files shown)


scanners/flutter/flutter.go
```diff
@@ -252,28 +251,8 @@ func (scanner *Scanner) Options() (models.OptionNode, models.Warnings, models.Ic
 		}
 		cfg := configName + "-" + testKey
 
-		if project.hasIosProject {
```
ofalvai, line 255: Remove this entire if block because the iOS related options are not needed anymore

lpusok (approved):

PR #228, opened 2022-02-08
Fix for duplicate Steps in react-native workflows

lpusok:
### Checklist
- [x] I've read and followed the [Contribution Guidelines](https://github.com/bitrise-steplib/.github/blob/main/CONTRIBUTING.md)
- [/] `step.yml` and `README.md` is updated with the changes (if needed)

### Version
Requires a *PATCH* [version update](https://semver.org/)

### Context
In case of multiple iOS schemes being present for react-native, duplicate _xcode-archive_ and _deploy-to-bitrise-io_ Steps were generated.

Resolves: https://bitrise.atlassian.net/browse/STEP-1742

### DIFF
--- a/scanners/reactnative/plain.go
+++ b/scanners/reactnative/plain.go
@@ -210,47 +210,47 @@ func (scanner *Scanner) configs(isPrivateRepo bool) (models.BitriseConfigMap, er
 	}
 
 	// ios cd
-	if scanner.iosScanner != nil {
-		for _, descriptor := range scanner.iosScanner.ConfigDescriptors {
-			if descriptor.MissingSharedSchemes {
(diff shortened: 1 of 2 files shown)


scanners/reactnative/plain.go
```diff
@@ -210,47 +210,47 @@ func (scanner *Scanner) configs(isPrivateRepo bool) (models.BitriseConfigMap, er
 	}
 
 	// ios cd
-	if scanner.iosScanner != nil {
-		for _, descriptor := range scanner.iosScanner.ConfigDescriptors {
```
lpusok, line 214: Removed this loop over iOS descriptors to prevent duplicate Steps added to the config. Use the first descriptor only.

PR #232, opened 2022-02-10
Seperated iOS Project parsing

lpusok:
### Context

To allow reusing native iOS project scanning for react-native workflows, extracted project parsing. This includes all filesystem operations.

Resolves: https://bitrise.atlassian.net/browse/STEP-1742

### Changes
* Extract project scanning from _GenerateOptions()_ into _ParseProjects()_. This way the options are generated from an intermediary representation _DetectResult_. This allows testing the project parsing independently and reusing it in the react-native scanner.
* Remved _Detect()_, merged with _ParseProjects()_

### DIFF
--- a/scanners/ios/ios.go
+++ b/scanners/ios/ios.go
@@ -2,37 +2,6 @@ package ios
 
 import "github.com/bitrise-io/bitrise-init/models"
 
-// Scheme is an Xcode project scheme or target
-type Scheme struct {
-	Name       string
-	Missing    bool
-	HasXCTests bool
-	HasAppClip bool
-
-	Icons       models.Icons
-	IconWarning string
-}
-
-// Project is an Xcode project on the filesystem
-type Project struct {
-	// Is it a standalone project or a workspace?
-	IsWorkspace    bool
-	IsPodWorkspace bool
-
-	RelPath string
-	Schemes []Scheme
-
-	// Carthage command to run: bootstrap/update
-	CarthageCommand string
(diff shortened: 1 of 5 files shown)


scanners/ios/ios.go
```diff
@@ -2,14 +2,47 @@ package ios
 
 import "github.com/bitrise-io/bitrise-init/models"
 
+// Scheme is an Xcode project scheme or target
+type Scheme struct {
```
ofalvai, line 6: These struct definitions are only used in `utility.go` so are shared between the macOS and iOS scanner. While I'm not a fan of shoveling everything into a file called `utility.go`, I think it would still be a better place than here in `ios.go`

scanners/ios/utility.go
```diff
@@ -549,37 +499,101 @@ func GenerateOptions(projectType XcodeProjectType, searchDir string, excludeAppI
...
+
+			iconsForAllProjects = append(iconsForAllProjects, scheme.Icons...)
+
+			iconIDs := []string{}
+			for _, icon := range scheme.Icons {
+				iconIDs = append(iconIDs, icon.Filename)
+			}
+
+			for _, exportMethod := range exportMethods {
+				// Wether app-clip export Step is added later depends on the used export method
```
ofalvai, line 580:
```suggestion
				// Whether app-clip export Step is added later depends on the used export method
```

ofalvai (approved):

PR #237, opened 2022-02-15
Detect unsupported tools in projects

ofalvai:
### Context

Detect unsupported tools/frameworks/platforms in the project directory to log them later.

### Changes

- Create `scanner/unknown.go` for everything related to scanning the repo for unsupported but recognized tools. This is not a scanner though, because it needs to run only after all other platform scanners fail.
- A common interface for the various tool detectors and 3 actual implementations:
  - `singleFileToolDetector` that looks for only a single file name
  - `multiFileToolDetector` that first looks for a primary file name, but can also find one of the optional files as a fallback
  - `kotlinMultiplatformDetector` because there is no distinct file name pattern for KMM, so it also looks at the file contents of certain files 🤷 
 - While traversing the folders recursively, the detectors also build a string representation of the file tree for logging purposes. Similar to the `tree` command, but a bit simpler.

### Decisions

- Maximum scan depth is 3
- A list of excluded directories is defined in code

### Out of scope for this PR

- Wiring this into the project scanner flow
- Logging of results

### DIFF
--- a/scanner/unknown.go
+++ b/scanner/unknown.go
@@ -28,10 +28,10 @@ type DetectionResult struct {
 }
 
 var UnknownToolDetectors = []UnknownToolDetector{
-	singleFileToolDetector{toolName: "Tuist", file: "Project.swift"},
-	singleFileToolDetector{toolName: "Xcodegen", file: "project.yml"},
-	singleFileToolDetector{toolName: "Swift Package Manager", file: "Package.swift"},
-	multiFileToolDetector{toolName: "Buck", primaryFile: "BUCK", optionalFiles: []string{".buckversion", ".buckconfig", ".buckjavaargs"}},
+	toolDetector{toolName: "Tuist", primaryFile: "Project.swift"},
+	toolDetector{toolName: "Xcodegen", primaryFile: "project.yml"},
+	toolDetector{toolName: "Swift Package Manager", primaryFile: "Package.swift"},
+	toolDetector{toolName: "Buck", primaryFile: "BUCK", optionalFiles: []string{".buckversion", ".buckconfig", ".buckjavaargs"}},
 	kotlinMultiplatformDetector{},
 }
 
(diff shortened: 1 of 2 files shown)


scanner/unknown.go
```diff
@@ -0,0 +1,209 @@
...
+	// DetectToolIn should search recursively in rootPath looking for the given tool
+	DetectToolIn(rootPath string) (DetectionResult, error)
+}
+
+type DetectionResult struct {
+	Detected    bool
+	ProjectTree string
+}
+
+var UnknownToolDetectors = []UnknownToolDetector{
```
tothszabi, line 30: Bazel is missing from the list.

scanner/unknown.go
```diff
@@ -0,0 +1,209 @@
...
+	".git",
+	".idea",
+	"node_modules",
+	"Pods",
+	"Carthage",
+	"CordovaLib",
+	"Framework",
+}
+
+type singleFileToolDetector struct {
```
tothszabi, line 48: The `singleFileToolDetector` and `multiFileToolDetector` implementations are too similar to each other which made me think we could unify them. Like a `singleFileToolDetector` can be represented with a `multiFileToolDetector` with a single required item and an empty optional items list. 

And if we do that then we can drop the first part of its name and just keep `toolDetector`.

ofalvai: That's a good idea actually!

scanner/unknown.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+var excludedDirs = []string{
+	".git",
+	".idea",
+	"node_modules",
+	"Pods",
+	"Carthage",
+	"CordovaLib",
+	"Framework",
```
ofalvai, line 45: @tothszabi is this the correct folder name for iOS frameworks?

scanner/unknown.go
ofalvai: Fixed!

tothszabi (approved):

PR #240, opened 2022-02-17
Log detected tools

ofalvai:
### Context

In #237 we added the ability to detect tools/frameworks that are not yet supported. This PR builds on that to run those tool detectors while scanning the project config and log any detected tool remotely.

### Changes

- Run the list of tool detectors in `scanner/result.go`, regardless of the scan result
- Use `analytics.LogInfo()` to log any detected tools
- Remove some debug log statements from the previous PR as I discovered that debug logs are enabled globally for the entire `bitrise-init` tool


shams-ahmed (approved):

tothszabi (approved):
Nice one

PR #242, opened 2022-02-25
[STEP-1085] CLI Cleanup

tothszabi:
This well remove all of the cli bits and convert `bitrise-init` to a package similar to `go-xcode` and the others we have.

This PR contains the following changes:
- Instead of having a global scanners list the package will now create fresh scanner instances on demand to avoid having shared state inside the package
- The cli code parts are removed
   - Integration tests do not use the ci built binary for testing
   - The `cli` and `version` folders are removed
   - Docker files and scripts are removed
- The unnecessary workflows are removed and only a single one for the ci is kept
- The `vendor` folder will be removed in a separate PR so it does not pollute this one
- Added a simple app in the `_manual-config` folder to still have an easy way to generate the manual config
   - It's dependencies are not vendored because this was the only way I could link to local files in the repo
- The Bitrise app (https://app.bitrise.io/app/cc3ccc097925deb0#/builds) will only keep the PR trigger and I will remove the push to master and tag one. Also the `create-release` and `release` workflows. But only just before I will merge this PR so you could still check them.


README.md
```diff
@@ -38,17 +19,33 @@ Please check available workflows in [bitrise.yml](./bitrise.yml).
 - create tag with the new version
 - test the generated release and its binaries
 
-__Update manual config on website__
+### Update manual config on website
+
+- Use the included go app to generate the manual configuration:
```
ofalvai, line 24: Nice little detail! 🤓

_manual-config/main.go
```diff
@@ -0,0 +1,41 @@
+package main
```
ofalvai, line 1: nit: I'd add some package-level documentation here about the purpose of this CLI tool

_tests/integration/helper/git.go
```diff
@@ -0,0 +1,22 @@
+package helper
+
+import (
+	"fmt"
+	"github.com/bitrise-io/go-utils/command/git"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func GitClone(t *testing.T, dir, uri string) {
```
ofalvai, line 10: nit: Now that we have an extracted git clone function that is used by every test, I'm wondering if there is anything we can do to speed up these slow test cases. Maybe executing `git clone --depth=1` in this function?

_tests/integration/manual_config_test.go
```diff
@@ -19,25 +21,25 @@ func TestManualConfig(t *testing.T) {
 	tmpDir, err := pathutil.NormalizedOSTempDirPath("__manual-config__")
```
ofalvai, line 21: nit: I would use `t.TempDir()`, it's automatically cleaned up after the test

ofalvai (approved):
Looks amazing, nice work! 💪
I'm okay with the trigger and workflow changes as well.

bitrise.yml
```diff
@@ -2,42 +2,6 @@ format_version: "7"
...
-        inputs:
-        - content: |-
-            #!/usr/bin/env bash
-            go get -u github.com/godrei/stepper
-
-            # explanation:
-            # 1. run stepper tool
-            # 2. extract relevant part
-            # https://stackoverflow.com/questions/7103531/how-to-get-the-part-of-a-file-after-the-first-line-that-matches-a-regular-expres
-            stepper stepLatests --steps-const-file=$BITRISE_SOURCE_DIR/steps/const.go | sed -n -e '/package steps/,$p' > stepper_output.go
```
lpusok, line 29: the stepper tool seems still useful, should we preserve it?

_tests/integration/manual_config_test.go
tothszabi: That is a great recommendation. I have checked the `pathutil.NormalizedOSTempDirPath` function and it is indeed using io function which explicitly states the folders need to be cleaned up manually. I have changed this everywhere.

_tests/integration/helper/git.go
tothszabi: That is a good one. I have tried to achieve this in the easiest way which does not need me to update the git package in the `go-utils`. The existing git functions are not flexible at all and I evaluated that I need to do a lot of things to achieve it. 

What I did was to manually inject the `--depth=1` option into the generated commands argument list before executing it. 🙈

Feel free to block it if you think it should be done in a proper way. 😄

bitrise.yml
tothszabi: I think you are right. I will add it back. 

(Although I have not used this stepper tool so far. Maybe I should have. 😄)

ofalvai (dismissed):

lpusok (approved):

PR #243, opened 2022-03-02
Handle Cocoapods parsing errors gracefully

ofalvai:
### Context

When `mergePodWorkspaceProjectMap()` returns with a non-nil error, the returned result type unconditionally overwrites the detected containers. This means that the detected containers will be empty and the project won't be recognized as an iOS project at all.

An error in the Cocoapods-related scanning shouldn't cause this, the scanner can still recognize the project's platform correctly, as well as find workspaces and schemes.

### DIFF
--- a/scanners/ios/utility.go
+++ b/scanners/ios/utility.go
@@ -66,9 +66,6 @@ const (
 	ExportMethodInputSummary = "The export method used to create an .app file in your builds, stored as an Environment Variable. You can change this at any time, or even create several .app files with different export methods in the same build."
 )
 
-// XCConfigContentInputKey ...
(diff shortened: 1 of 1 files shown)


scanners/ios/utility.go
```diff
@@ -66,9 +66,6 @@ const (
 	ExportMethodInputSummary = "The export method used to create an .app file in your builds, stored as an Environment Variable. You can change this at any time, or even create several .app files with different export methods in the same build."
 )
 
-// XCConfigContentInputKey ...
```
ofalvai, line 69: Unused variable

PR #249, opened 2023-02-14
New cache steps for Android & iOS

ofalvai:
### Changes

- Functional
  - Replace Cache Pull/Push steps in Android workflows with Restore/Save Gradle cache
  - Replace Cache Pull/Push steps in iOS workflows with Restore/Save Cocoapods/Carthage cache
  - Remove cache steps from`deploy` workflows. Deploying to stores should probably run a clean build to maximize reproducibility and stability.
- Refactors and cleanups
  - Remove dummy comments on exported methods
  - Test cleanups

### Out of scope

Adding SPM cache steps. This needs new detection logic, but this PR is already large, so it will come in a future PR.

Other platforms are also out of scope for this PR.

### DIFF
--- a/_tests/integration/fastlane_test.go
+++ b/_tests/integration/fastlane_test.go
@@ -128,6 +128,7 @@ configs:
               inputs:
               - lane: $FASTLANE_LANE
               - work_dir: $FASTLANE_WORK_DIR
+              - enable_cache: "no"
           - deploy-to-bitrise-io@%s: {}
   ios:
     ios-test-config: |
--- a/_tests/integration/fastlane_test.go
+++ b/_tests/integration/fastlane_test.go
@@ -153,12 +153,14 @@ configs:
               - project_path: $BITRISE_PROJECT_PATH
               - scheme: $BITRISE_SCHEME
               - test_repetition_mode: retry_on_failure
+              - cache_level: none
           - xcode-archive@%s:
               inputs:
               - project_path: $BITRISE_PROJECT_PATH
               - scheme: $BITRISE_SCHEME
               - distribution_method: $BITRISE_DISTRIBUTION_METHOD
               - automatic_code_signing: api-key
+              - cache_level: none
           - deploy-to-bitrise-io@%s: {}
         primary:
           description: |
@@ -174,6 +176,7 @@ configs:
               - project_path: $BITRISE_PROJECT_PATH
               - scheme: $BITRISE_SCHEME
               - test_repetition_mode: retry_on_failure
+              - cache_level: none
           - deploy-to-bitrise-io@%s: {}
 warnings:
   fastlane: []
--- a/_tests/integration/fastlane_test.go
+++ b/_tests/integration/fastlane_test.go
@@ -10,8 +10,6 @@ import (
 )
 
 func TestFastlane(t *testing.T) {
-	tmpDir := t.TempDir()
-
 	var testCases = []helper.TestCase{
 		{
 			"fastlane",
@@ -22,7 +20,7 @@ func TestFastlane(t *testing.T) {
 		},
 	}
 
-	helper.Execute(t, tmpDir, testCases)
+	helper.Execute(t, testCases)
 }
 
 // Expected results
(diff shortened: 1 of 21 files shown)


steps/steps.go
```diff
@@ -37,7 +37,6 @@ func stepListItem(stepIDComposite, title, runIf string, inputs ...envmanModels.E
 	}
 }
 
-// DefaultPrepareStepList ...
```
tothszabi, line 40: The removal of these comments indicates that we are not running the same CI stack as we do for our steps. I would probably keep it for now and instead hook this up to the regular workflow. (But obviously not in this PR.)

_tests/integration/fastlane_test.go
```diff
@@ -41,17 +39,13 @@ var fastlaneVersions = []interface{}{
 	models.FormatVersion,
 	steps.ActivateSSHKeyVersion,
 	steps.GitCloneVersion,
-	steps.CachePullVersion,
```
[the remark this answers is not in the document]
lpusok: Android etc. Steps also have legacy cache enabled by deafult, good point. Propably they need cache input set to off.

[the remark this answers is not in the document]
ofalvai: Good point about turning off legacy caching in the relevant steps, I'll update this soon.

SPM dependencies can be present in any of the Apple-related project types, and I'm working right now on detecting them, so I'd rather not update all the tests just to update them again when the SPM cache steps can be added to the workflows.

steps/steps.go
ofalvai: We actually do! I made this PR a few days ago: https://github.com/bitrise-io/bitrise-init/pull/248

I wanted to get rid of these dummy comments, which are enforced by the old `golint` project. That is replaced with `golangci-lint`, the way we lint step repos.

steps/steps.go
```diff
@@ -82,11 +79,10 @@ func DefaultDeployStepList(isIncludeCache bool) []bitriseModels.StepListItemMode
 	return stepList
 }
 
-// DefaultDeployStepListV2 ...
-func DefaultDeployStepListV2(shouldIncludeCache bool) []bitriseModels.StepListItemModel {
+func DefaultDeployStepListV2(shouldIncludeLegacyCache bool) []bitriseModels.StepListItemModel {
 	stepList := []bitriseModels.StepListItemModel{}
 
-	if shouldIncludeCache {
+	if shouldIncludeLegacyCache {
```
lpusok, line 85: Maybe this parameter can be removed completely?

ofalvai: Actually, yes. I wanted to clean this up after updating all the default workflows, but I just checked now and every call is `false` in this PR.

ofalvai: Actually, no 😅 Although this parameter is unused, its pair in the "default end of workflows steps" function is still used by the Flutter scanner to add the `cache-push` step. I'll clean this up once I change the Flutter workflows.

_tests/integration/fastlane_test.go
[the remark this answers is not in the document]
ofalvai: Take a look now

_tests/integration/fastlane_test.go
```diff
@@ -128,6 +128,7 @@ configs:
               inputs:
               - lane: $FASTLANE_LANE
               - work_dir: $FASTLANE_WORK_DIR
+              - enable_cache: "no"
```
lpusok, line 131: Could the iOS + Android cache steps be added here?

lpusok:
Cocopods-install (and yarn, carthage possibly) also has legacy cache inputs that may need to be disabled

lpusok (approved):

_tests/integration/fastlane_test.go
ofalvai: After a deeper look, I don't think it's a good idea. This workflow is offered when the user selects a raw Fastlane workflow (even though the scanner recognizes the project as iOS/Android too). So I don't see how we could select here one of the Gradle/SPM/Carthage/Cocoapods cache steps.

I also wanted to include an empty generic cache step pair, but then realized that they have required inputs, I can't leave the cache key unfilled.

PR #250, opened 2023-02-16
Add NPM cache to JS platform workflows

ofalvai:
### Changes

Add NPM cache steps to:

- React Native and Expo workflows
- Cordova workflows
- Ionic workflows

### Notes

NPM and Yarn steps have a legacy caching input, but both of them are off by default, so this PR doesn't add those inputs to the workflow.

### DIFF
--- a/_tests/integration/cordova_test.go
+++ b/_tests/integration/cordova_test.go
@@ -47,8 +47,10 @@ var sampleAppsCordovaWithJasmineVersions = []interface{}{
 	steps.ActivateSSHKeyVersion,
 	steps.GitCloneVersion,
 	steps.ScriptVersion,
+	steps.CacheRestoreNPMVersion,
 	steps.NpmVersion,
 	steps.JasmineTestRunnerVersion,
+	steps.CacheSaveNPMVersion,
 	steps.DeployToBitriseIoVersion,
 }
 
@@ -99,10 +101,12 @@ configs:
           - git-clone@%s: {}
           - script@%s:
               title: Do anything with Script step
+          - restore-npm-cache@%s: {}
           - npm@%s:
               inputs:
               - command: install
           - jasmine-runner@%s: {}
+          - save-npm-cache@%s: {}
           - deploy-to-bitrise-io@%s: {}
 warnings:
   cordova: []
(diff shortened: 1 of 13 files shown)


ofalvai:
Current dependencies on/for this PR:
* master
  * **PR #250** <a href="https://app.graphite.dev/github/pr/bitrise-io/bitrise-init/250" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/bitrise-io/bitrise-init/250?utm_source=stack-comment).

_tests/integration/cordova_test.go
```diff
@@ -99,10 +101,12 @@ configs:
           - git-clone@%s: {}
           - script@%s:
               title: Do anything with Script step
+          - restore-npm-cache@%s: {}
```
godrei, line 104: Wouldn't it be better if cache-restore directly follows git-clone?

steps/const.go
```diff
@@ -42,13 +42,17 @@ const (
 	CacheRestoreCocoapodsVersion = "1"
 	CacheRestoreCarthageID       = "restore-carthage-cache"
 	CacheRestoreCarthageVersion  = "1"
+	CacheRestoreNPMID            = "restore-npm-cache"
```
godrei, line 45: Do we still use the old cache steps in the default workflows?

steps/const.go
```diff
@@ -42,13 +42,17 @@ const (
...
+	CacheRestoreNPMVersion       = "1"
```
godrei, line 46: NOTE: Do we want to update all the steps to the latest version in the default workflows (for example git-clone has a new major version)

ofalvai: Good idea, but let's not do that in this PR. I'll do it in a separate PR before deploying the new `bitrise-init`

steps/const.go
ofalvai: Yes, I didn't touch the Flutter workflows yet. I'll clean up the old cache steps once I'm done with all platforms.

_tests/integration/cordova_test.go
ofalvai: Wouldn't it be even better to remove the script step? 😅 AFAIK we agreed to remove all of them, Cordova seems to be a leftover.

ofalvai: Fixed

godrei (approved):

ofalvai:
⏳ This pull request is set to merge as part of a Graphite merge job
Job ID: 6dQCJ6udRdY0ZfxVWVla.
See details on [graphite.dev](https://app.graphite.dev/github/pr/bitrise-io/bitrise-init/250)

PR #252, opened 2023-02-17
Flutter cache steps, cleanup legacy cache steps

ofalvai:
### Changes

- Replace legacy cache steps with new steps in Flutter workflows
- Clean up the last remaining references to the legacy cache steps
- Clean up the duplicated prepare and deploy functions as the old ones are not used anymore

PR #254, opened 2023-03-20
Handle public repos in default workflows correctly

ofalvai:
### Context

Fix a regression caused in #252: when generating the default configs where the `isPrivateRepository` information is not available, we default to adding the `activate-ssh-key` step to those default workflows. This is a good practice, but it breaks public repos configured with the default configs because this step requires the `SSH_RSA_PRIVATE_KEY` env var set up by the website. But because we create the default configs, the website doesn't set up any SSH keys and doesn't expose this variable.

In the past, we used to include the `activate-ssh-key` unconditionally and create a `run_if` condition that checks the presence of the required env var. It's probably still a better idea to only include the SSH step if the repo is private and simplify the workflow for public repos.

### Changes

The basic idea is that we should keep the existing logic that only adds the activate step if the repo is private. But this is not a binary choice, there is also the case when the repo's access is unknown, and in this case we should still add the step, but with the old run condition.

In more detail:
- Add a new enum called `models.RepoAccess` with the 3 possible states
- Change the scanner interface's config method from `Configs(isPrivateRepository bool) (models.BitriseConfigMap, error)` to `Configs(repoAccess models.RepoAccess) (models.BitriseConfigMap, error)`
- Update the SSH step generator code


models/configbuilder_test.go
```diff
@@ -3,13 +3,16 @@ package models
 import (
 	"testing"
 
-	"github.com/bitrise-io/bitrise-init/steps"
```
ofalvai, line 6: This caused an import cycle after adding `models.RepoAccess` and depending on it from `steps`. I think this test code shouldn't depend on the `steps` package anyway, so I updated the test setup code.

steps/steps.go
```diff
@@ -39,8 +40,20 @@ func stepListItem(stepIDComposite, title, runIf string, inputs ...envmanModels.E
 func DefaultPrepareStepList(params PrepareListParams) []bitriseModels.StepListItemModel {
 	stepList := []bitriseModels.StepListItemModel{}
 
-	if params.ShouldIncludeActivateSSH {
-		stepList = append(stepList, ActivateSSHKeyStepListItem(""))
+	switch params.RepoAccess {
+	case models.RepoAccessPublic: // No SSH key setup needed
```
tothszabi, line 44: Would it make sense to remove this line and just leave the same comment (`// No SSH key setup needed for public repos`) on top of the switch statement? It tricked me because first I though that the empty case will fall through to the second and I had to google how Go behaves to realise that will not happen. But it made me do this extra work which I think others might go through and could be avoided. 😄

ofalvai: We have the `exhaustive` linter enabled in this project (for good reason, I think it's very useful!), so I don't want to completely remove this case, but I can wrap the comment in curly braces, take a look now.

tothszabi (approved):

PR #257, opened 2023-05-18
Log Flutter project SDK constraints

godrei:
This PR updates the Flutter scanner to send analytic data about the given Flutter project's SDK constraints.
The plan is to use this information for improving the Flutter Installer step and the preinstalled Flutter versions.

Related task: https://bitrise.atlassian.net/browse/BIVS-1851

Changes:
- Introduce `flutterproject` package for reading the Flutter and Dart SDK constraints of the given Flutter project
  - The entry point to the functionality is the `Project.FlutterAndDartSDKVersions`, it runs SDKVersionsReaders and collects the detected SDK versions
  - SDKVersionsReaders are implemented for fvm, asdf, pubspec.lock and pubspec.yaml
- Introduce `tracker` packager for tracking the SDK constraints
- Non-functional: pointer receiver for flutter scanner functions


scanners/flutter/flutter.go
```diff
@@ -225,15 +230,20 @@ projects:
...
-	if len(scanner.projects) == 0 {
-		return false, nil
+	for _, project := range scanner.projects {
+		proj := flutterproject.New(project.path, flutterproject.NewFileOpener())
+		sdkVersions, err := proj.FlutterAndDartSDKVersions()
+		if err == nil {
+			scanner.tracker.LogSDKVersions(sdkVersions)
+		}
 	}
+	scanner.tracker.Wait()
```
vshah23, line 240: Is it necessary to wait here?

Why not make logging analytics a fire and forget operation or move the call to `Wait` to just before exiting the program?

scanners/flutter/flutterproject/fileopener.go
```diff
@@ -0,0 +1,25 @@
+package flutterproject
+
+import (
+	"io"
+	"os"
+)
+
+type FileOpener interface {
```
vshah23, line 8: I immediately feel like this should be in `go-utils`. Actually, I'm surprised it's not already there.

vshah23 (approved):

scanners/flutter/flutterproject/pubspec.go
```diff
@@ -0,0 +1,72 @@
...
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+
+	var dartVersion *VersionConstraint
+	if dartVersionStr != "" {
+		dartVersion, err = NewVersionConstraint(dartVersionStr, PubspecVersionSource)
+		if err != nil {
+			return nil, nil, err
```
vshah23, line 50: Not every Flutter project will have the Dart SDK version specified in the pubspec and I think even nowadays the Flutter version is not required.

Is this really an error scenario?

See our demo project: https://github.com/bitrise-io/flutter-demo-app/blob/main/pubspec.yaml#L21

scanners/flutter/flutterproject/pubspec_lock.go
```diff
@@ -0,0 +1,72 @@
...
+	flutterVersionStr, dartVersionStr, err := parsePubspecLockSDKVersions(f)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var flutterVersion *VersionConstraint
+	if flutterVersionStr != "" {
+		flutterVersion, err = NewVersionConstraint(flutterVersionStr, PubspecLockVersionSource)
+		if err != nil {
+			return nil, nil, err
```
vshah23, line 42: Same as `pubspec.yaml`, is it really an error scenario if one of the SDK versions is missing?

From our demo app: https://github.com/bitrise-io/flutter-demo-app/blob/main/pubspec.lock

scanners/flutter/flutterproject/pubspec.go
vshah23: Ah I missed the part where we're checking if the string is not empty first 🤦‍♂️

scanners/flutter/flutterproject/pubspec_lock.go
vshah23: Noticed you already cover this in the unit test and I missed the check on line 39 in my initial review.

scanners/flutter/flutter.go
godrei: I just wanted to make sure that we log the data before the scanner gets freed up.

scanners/flutter/flutterproject/fileopener.go
godrei: Here is the PR: https://github.com/bitrise-io/go-utils/pull/182

godrei: The current FileOpener is a bit specific to this use case:
- instead of the original return value of os.Open, which is an os.File, it returns an io.Reader
- instead of the original os.Open, it doesn't produce an error if the file doesn't exist

I wouldn't move this specific version to the go-utils and the more generic solution wouldn't help in this code much, because I still need to implement the same FileOpener, but instead of calling os.Open it would call FileManager.Open.

Because of this, I think we should leave this code as it is, but I think we can still merge the go-utils PR with the generic FileManager.Open.

godrei:
Flutter SDK tracking functionality has moved to the Flutter Installer Step in this PR: https://github.com/bitrise-steplib/bitrise-step-flutter-installer/pull/32

PR #261, opened 2023-06-29
[CI-1640] Add SPM project support

tothszabi:
This PR adds pure SPM project support to the iOS scanner. The Mac version will come in a separate PR.

A couple of implementation details:

1.) 
The solution uses the built in swift toolchain to get a json output of the project setup:
```
swift package dump-package
```

2.)
The most common use case of the top SPM projects is to have single product in an SPM project which most of the time is a library product. But you can easily define more than one product in your project. 

Lets take this example:
```
...
let package = Package(
    name: "CoolFeature",
    products: [
        .library(name: "CoolFeature",  targets: ["CoolFeature"]),
    ],
    ...
)
```
When you initialise an empty package with Swift then it sets the package and product name to the same value similar to the example above. In this case Xcode will generate a single scheme called `CoolFeature`.

But in a multi product example:
```
...
let package = Package(
    name: "CoolFeature",
    products: [
        .library(name: "CoolFeature",  targets: ["CoolFeature"]),
        .library(name: "CoolFeature Static", type: .static, targets: ["CoolFeature"]),
        .library(name: "CoolFeatureDynamic", type: .dynamic, targets: ["CoolFeature"]),
    ],
    ...
)
```
Xcode will create four schemes. Three will be for the individual products and one for the package. But it is clear that the package and the first library scheme name will be in conflict as they have the same name. SPM solved this by always automatically appending a `-Package` string to the scheme created from the package name. 

The other interesting fact is that only the scheme created from the package will contains all of the test targets. So the other targets are not that useful than this single main one. So the SPM part of the scanner only looks for this main scheme.

3.)
SPM projects cannot really have a meaningful deploy workflow without user input. There are just a lot of different ways a library can be distributed. So we will not provide a deploy workflow for SPM projects.

### DIFF
--- a/scanners/ios/spmproject.go
+++ b/scanners/ios/spmproject.go
@@ -28,11 +28,16 @@ type spmTarget struct {
 	Type string `json:"type"`
 }
 
+// We do not need the properties because we only the check the dependency existence.
+type spmDependency struct {
+}
+
 type spmProject struct {
-	Name      string        `json:"name"`
-	Platforms []spmPlatform `json:"platforms"`
-	Products  []spmProduct  `json:"products"`
-	Targets   []spmTarget   `json:"targets"`
+	Name         string          `json:"name"`
+	Platforms    []spmPlatform   `json:"platforms"`
+	Products     []spmProduct    `json:"products"`
+	Targets      []spmTarget     `json:"targets"`
+	Dependencies []spmDependency `json:"dependencies"`
 }
 
 func ParseSPMProject(projectType XcodeProjectType, searchDir string) (DetectResult, error) {
@@ -74,9 +79,10 @@ func ParseSPMProject(projectType XcodeProjectType, searchDir string) (DetectResu
 		Warnings:        nil,
 		Schemes:         []Scheme{scheme},
 	}
+	hasDependencies := 0 < len(proj.Dependencies)
 	result := DetectResult{
 		Projects:           []Project{project},
-		HasSPMDependencies: true,
+		HasSPMDependencies: hasDependencies,
 		Warnings:           nil,
 	}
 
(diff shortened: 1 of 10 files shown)


scanners/ios/utility.go
```diff
@@ -503,7 +528,14 @@ func GenerateOptions(projectType XcodeProjectType, result DetectResult) (models.
...
 			for _, exportMethod := range exportMethods {
 				// Whether app-clip export Step is added later depends on the used export method
-				configDescriptor := NewConfigDescriptor(project.IsPodWorkspace, project.CarthageCommand, scheme.HasXCTests, scheme.HasAppClip, result.HasSPMDependencies, exportMethod, scheme.Missing)
+				configDescriptor := NewConfigDescriptor(
+					project.IsPodWorkspace,
+					project.CarthageCommand,
+					scheme.HasXCTests,
+					scheme.HasAppClip,
+					result.HasSPMDependencies,
+					project.IsSPMProject,
```
godrei, line 537: here we could implicitly use `false` instead of project.IsSPMProject, to signal that in this case, it isn't an SPM project for sure

scanners/ios/utility.go
```diff
@@ -640,6 +677,7 @@ func GenerateDefaultConfig(projectType XcodeProjectType) (models.BitriseConfigMa
 		true,
 		false,
 		true,
+		false,
```
godrei, line 680: I think this way we will provide the default ios workflows for an SPM project when the manual config is selected.
This config contains steps, which are not valid for SPM and a deploy workflow.

scanners/reactnative/plain.go
```diff
@@ -71,7 +71,7 @@ func generateIOSOptions(result ios.DetectResult, hasAndroid, hasTests, hasYarnLo
 			schemeOption.AddOption(scheme.Name, exportMethodOption)
 
 			for _, exportMethod := range ios.IosExportMethods {
-				iosConfig := ios.NewConfigDescriptor(project.IsPodWorkspace, project.CarthageCommand, scheme.HasXCTests, scheme.HasAppClip, result.HasSPMDependencies, exportMethod, scheme.Missing)
+				iosConfig := ios.NewConfigDescriptor(project.IsPodWorkspace, project.CarthageCommand, scheme.HasXCTests, scheme.HasAppClip, result.HasSPMDependencies, project.IsSPMProject, exportMethod, scheme.Missing)
```
godrei, line 74: Isn't project.IsSPMProject always false in this case?

scanners/ios/spmproject.go
```diff
@@ -28,11 +28,16 @@ type spmTarget struct {
 	Type string `json:"type"`
 }
 
+// We do not need the properties because we only the check the dependency existence.
```
godrei, line 31: I think some words are missing from this sentence.

godrei (approved):

PR #263, opened 2023-07-12
[CI-1641] Kotlin build script support

tothszabi:
The generated deploy workflow fails out of the box with Kotlin build scripts. The reason is that it uses a hardcoded value which needs to be dynamic (and in Kotlin's case be `build.gradle.kts`):
https://github.com/bitrise-io/bitrise-init/blob/7c671dd33404fea833bf77854dcef3c95efc10cd/scanners/android/utility.go#L180

This PR updates the scanner to store if the build scripts are Kotlin based or not for every project it finds. This information will be used later at the config generation phase. 

At the moment the scanner always generated the same config for every Android project. Now it will use the previously stored knowledge about the build files. If it has at least one project with Groovy based scripts then it will generate the `android-config` and if the project is Kotlin script based then it will generate the `android-config-kts`. If it has multiple project of both types then it will generate both. 

The only difference between `android-config` and `android-config-kts` is the single hardcoded path value value to the build file.


scanners/android/android.go
```diff
@@ -129,6 +137,15 @@ func parseProjects(searchDir string, projectRoots []string) ([]Project, error) {
 	return projects, nil
 }
 
+func usesKotlinBuildScripts(projectRoot string) bool {
```
ofalvai, line 140: Is this going to work when the Gradle project is in a subfolder? For example, React Native projects put the whole Gradle project in `./android`, or a monorepo where the Android client is in a subfolder.

scanners/android/utility.go
```diff
@@ -176,8 +178,13 @@ func (scanner *Scanner) generateConfigBuilder(repoAccess models.RepoAccess) mode
 		envmanModels.EnvironmentItemModel{GradlewPathInputKey: gradlewPath},
 	))
 
+	basePath := filepath.Join(projectLocationEnv, moduleEnv)
+	path := filepath.Join(basePath, "build.gradle")
+	if useKotlinBuildScript {
+		path = filepath.Join(basePath, "build.gradle.kts")
```
ofalvai, line 184: nit: `android.go` already defines a constant for this filename

scanners/android/android.go
```diff
@@ -129,6 +137,15 @@ func parseProjects(searchDir string, projectRoots []string) ([]Project, error) {
...
+	return fileExists(filepath.Join(projectRoot, gradleKotlinBuildFile)) && fileExists(filepath.Join(projectRoot, gradleKotlinSettingsFile))
+}
+
+func fileExists(path string) bool {
```
godrei, line 144: NOTE: in the spmproject.go we also have a `fileExists`, shouldn't we share these codes?

scanners/android/android.go
```diff
@@ -172,26 +193,63 @@ func (scanner *Scanner) DefaultOptions() models.OptionNode {
...
+			useKotlinScript: true,
+		})
+	}
+	return params
 }
 
 // DefaultConfigs ...
 func (scanner *Scanner) DefaultConfigs() (models.BitriseConfigMap, error) {
-	configBuilder := scanner.generateConfigBuilder(models.RepoAccessUnknown)
+	configBuilder := scanner.generateConfigBuilder(models.RepoAccessUnknown, false)
```
godrei, line 252: Do we intentionally not add support for Kotlin build script-based android projects when manual config is selected?

scanners/android/android.go
tothszabi: I have checked and it will work. The Android scanner uses this method to traverse the directory layout from a starting folder:
https://github.com/bitrise-io/bitrise-init/blob/c9266f2c255a700a64a8e9a6352785fe747e169c/scanners/android/utility.go#L88

And it will find all of possible Android projects. From here it will filter down further by only using those potential project roots where there is a gradlew wrapper:
https://github.com/bitrise-io/bitrise-init/blob/8ea7233cb5f2763d4f7bcaebdc4be240acd5b23e/scanners/android/android.go#L104

And based on my research there is always a `build.gradle` or `build.gradle.kts` file (and the settings equivalent) next to the wrapper. Is that a correct assumption?

scanners/android/utility.go
tothszabi: I created that constant and forgot about it. 😄

scanners/android/android.go
tothszabi: I did it on purpose because that means we need to ask the user an extra question about their usage of Kotlin build scripts. Right know we ask the project location, module and variant. I was thinking that I do not want to add an extra question which lengthens the ANA flow for fixing that bug. 

It is not a hard a opinion and I can be persuaded to add it to the manual config too. It is not much work because it only needs an extra option here:
https://github.com/bitrise-io/bitrise-init/blob/8ea7233cb5f2763d4f7bcaebdc4be240acd5b23e/scanners/android/android.go#L181-L192

And making sure that both Android configs are generated for as part of the default config.

What is your opinion? Would you add it to the manual config?

scanners/android/android.go
ofalvai: Yes, that's correct, I was thinking about the pattern when the entire Gradle project is in a subfolder. But I'm glad it's going to work!

ofalvai (approved):

scanners/android/android.go
godrei: I think by default we should keep feature parity in automatic and manual configs. If we don't have other decision I would add it to the manual configs, but it can go into a separate PR.

godrei (dismissed):

scanners/android/android.go
tothszabi: You convinced me and added the manual config part in this PR.

godrei (approved):

PR #265, opened 2023-07-12
Update ios and android workflows

godrei:
### Version

### Context

This PR updates the generated iOS and Android workflow IDs, summaries, and descriptions.

The new workflow IDs better reflect what the workflow actually does:

For iOS:
- `primary` -> `run_tests` or `build` (if the project doesn't have tests)
- `deploy` -> `archive_and_export_app`

For Android:
- `primary` -> `run_tests`
- `deploy ` -> `build_apk`

New summaries talk about the main outcome, while descriptions explain high-level what happens in the workflows.

Resolves: https://bitrise.atlassian.net/browse/BIVS-2004

The website needs to be updated first, to handle the new workflow IDs: https://github.com/bitrise-io/bitrise-website/pull/10862

### Changes

- Update iOS and Android workflow IDs, summaries, and descriptions


scanners/ios/workflow.go
```diff
@@ -167,14 +219,22 @@ func addSharedTeardownSteps(workflow models.WorkflowID, params workflowSetupPara
...
 		return
 	}
 
 	configBuilder.SetWorkflowDescriptionTo(workflow, description)
 }
 
+func addSummary(workflow models.WorkflowID, configBuilder *models.ConfigBuilderModel, summary string) {
+	if summary == "" {
+		return
+	}
```
tothszabi, line 233: Does this empty string check needed? The same for the `addDescription` method too. 

In the old `addDescription` method implementation it was guarding against the macOS project types. I can see that  the builder is generated here:
https://github.com/bitrise-io/bitrise-init/blob/02e510ae4f764d344d615a24b87634e9ff5e5ecc/scanners/ios/utility.go#L602

And it has an empty summary and description when it is created. And also a separate builder is created when invoked from the macOS scanner. So to it seems like there will not be a possibility when the builder already has a summary and we could accidentally override it with an empty string.

scanners/ios/workflow.go
```diff
@@ -15,24 +15,29 @@ const (
 	AutomaticCodeSigningValue             = "api-key"
 	CacheLevelKey                         = "cache_level"
 	CacheLevelNone                        = "none"
-)
 
-const primaryTestDescription = `The workflow executes the tests. The *retry_on_failure* test repetition mode is enabled.`
+	// test workflow
+	primaryWorkflowID = "primary"
```
tothszabi, line 20: I thought almost though that the existence of the `primaryWorkflowID` and `deployWorkflowID` constants is an error but then realised that they are still used with macOS. I would propose to completely remove these and update the Mac scanner as well. I have two options for it:
1.)
Just reuse the same `run_tests`, `build` and `archive_and_export_app` because those workflows do almost the same thing. Expect the `build` one because we do not have a build step on Mac so that one does nothing useful. But still the intention would be the same so I can justify having the same name.

2.)
The second option is to also reuse the same description too. The `run_tests` and `build` can be reused as they are. The `archive_and_export_app` needs a separate one because instead of an ipa we have an exported app package.

scanners/ios/workflow.go
```diff
@@ -15,24 +15,29 @@ const (
...
+	buildWorkflowDescription = "The workflow will first clone your Git repository, cache and install your project’s dependencies if any and build your project."
 
-const primaryCommonDescription = `Next steps:
-- Check out [Getting started with iOS apps](https://devcenter.bitrise.io/en/getting-started/getting-started-with-ios-apps.html).
-`
+	// deploy workflow
+	deployWorkflowID = "deploy"
 
-const deployDescription = `The workflow tests, builds and deploys the app using *Deploy to bitrise.io* step.
+	archiveAndExportWorkflowID = "archive_and_export_app"
```
tothszabi, line 33: Was there a discussion about the format of name? I cannot describe why but to me the name with `_` is harder to read then for example with `-`. You also used dashes in your PR name instead of underscores. And a couple of apps also replace spaces with dashes. I can only think about the Github desktop app but I am sure I have encountered more. 😄

So my question is what if we use dashes instead of underscores to separate the words in the new workflow names?

scanners/android/android.go
```diff
@@ -8,9 +8,47 @@ import (
...
+	ConfigName        = "android-config"
+	DefaultConfigName = "default-android-config"
+
+	testsWorkflowID         = "run_tests"
+	testsWorkflowSummary    = "Run your Android unit tests and get the test report."
+	testWorkflowDescription = "The workflow will first clone your Git repository, cache your Gradle dependencies, install Android tools, run your Android unit tests and save the test report."
+
+	buildWorkflowID          = "build_apk"
+	buildWorkflowSummary     = "Run your Android unit tests and create an APK file to install your app on a device or share it with your team."
+	buildWorkflowDescription = "The workflow will first clone your Git repository, install Android tools, set the project’s version code based on the build number, run Android lint and unit tests, build the project’s APK file and save it."
```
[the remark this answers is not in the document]
godrei: The old descriptions were linked to 'getting started' and 'code signing' related pages. Along with the first build experience updates, we will show these guides (and more) on the new Workflows and Pipelines page in a dedicated UI element.

scanners/ios/workflow.go
godrei: For now, the aim is updating the ios and android workflows only.

scanners/ios/workflow.go
godrei: Updated in https://github.com/bitrise-io/bitrise-init/pull/265/commits/22bdb98cc509306a312db9d1426f9928053d6b9c

scanners/android/android.go
```diff
@@ -40,6 +78,171 @@ func (scanner *Scanner) DetectPlatform(searchDir string) (_ bool, err error) {
...
+	configBuilder.AppendStepListItemsTo(buildWorkflowID, steps.SignAPKStepListItem())
+	configBuilder.AppendStepListItemsTo(buildWorkflowID, steps.DefaultDeployStepList()...)
+
+	configBuilder.SetWorkflowDescriptionTo(buildWorkflowID, buildWorkflowDescription)
+	configBuilder.SetWorkflowSummaryTo(buildWorkflowID, buildWorkflowSummary)
+
+	return *configBuilder
+}
+
 func detect(searchDir string) ([]Project, error) {
```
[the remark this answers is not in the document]
godrei: Updated in https://github.com/bitrise-io/bitrise-init/pull/265/commits/649037f1667ead79f2718a1a037aabb472aa6a1a

scanners/ios/utility.go
```diff
@@ -614,7 +614,7 @@ func GenerateConfigBuilder(
 		exportMethod:         exportMethod,
 	}
 
-	createPrimaryWorkflow(params)
+	createTestWorkflow(params)
```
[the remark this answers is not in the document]
godrei: updated in https://github.com/bitrise-io/bitrise-init/pull/265/commits/a8728fd3b43e5cec47e942ca09d8da66cad6cc3a

tothszabi (dismissed):

tothszabi (approved):

PR #278, opened 2025-01-23
Use xcode-test@6

ofalvai:
### Version

Requires a *MAJOR/MINOR/PATCH* [version update](https://semver.org/)

### Context

https://github.com/bitrise-steplib/steps-xcode-test/releases/tag/6.0.0 was released more than a month ago, it's good enough now for the default.

### Changes

Also had to upgrade Go because CI was broken (it was using the Xcode 14.3 stack)


scanners/ios/podfile.go
```diff
@@ -33,6 +33,9 @@ func (podfileParser podfileParser) getTargetDefinitionProjectMap(cocoapodsVersio
 gem 'cocoapods'%s
 gem 'cocoapods-core'
 gem 'json'
+# warning: bigdecimal was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.4.0.
```
ofalvai, line 36: This is printed to stdout, so it broke the JSON parser below.

lpusok (approved):

PR #279, opened 2025-03-21
[CI-4595] Add parallelized test runner Pipeline example to iOS configs

AndrasEszes:
### Version

### Context

This PR introduces Pipelines and parallel test runners (test sharding) in iOS project templates. The original workflows remain unchanged, but with this, our customers gain insights into these useful features.

Resolves: https://bitrise.atlassian.net/browse/CI-4510

### Changes

Introduce two workflows for each iOS project template that contains tests:
- build_for_testing  
- test_without_building  

Introduce a new Pipeline with parallel test running that includes these two workflows:
- run_tests_parallel


_tests/integration/fastlane_test.go
tothszabi, line 1: Do we want to exclude fastlane? Because for fastlane the users need to write a fastlane code to actually build their projects so they will not use our steps. So giving them a pipeline with Bitrsie steps might not be useful.

_tests/integration/ios_test.go
```diff
@@ -190,6 +216,26 @@ configs:
...
           - deploy-to-bitrise-io@%s: {}
+        build_for_testing:
+          steps:
+          - activate-ssh-key@%s: {}
+          - git-clone@%s: {}
+          - xcode-build-for-test@%s:
+              inputs:
+              - project_path: $BITRISE_PROJECT_PATH
+              - scheme: $BITRISE_SCHEME
+              - destination: platform=iOS Simulator,name=iPhone 8 Plus,OS=latest
```
tothszabi, line 227: Let's use `generic/platform=iOS Simulator` as the destination input here.

_tests/integration/ios_test.go
```diff
@@ -204,6 +250,13 @@ configs:
               - test_repetition_mode: retry_on_failure
               - cache_level: none
           - deploy-to-bitrise-io@%s: {}
+        test_without_building:
+          steps:
+          - pull-intermediate-files@%s: {}
+          - xcode-test-without-building@%s:
+              inputs:
+              - only_testing: $BITRISE_TEST_SHARDS_PATH/$BITRISE_IO_PARALLEL_INDEX
+              - destination: platform=iOS Simulator,name=iPhone 8 Plus,OS=latest
```
tothszabi, line 259: Let's remove the destination input and rely on the step's default input value. We need to update this from time to time and it is enough if we need to update it at one place only.

_tests/integration/ios_test.go
```diff
@@ -190,6 +216,26 @@ configs:
...
+          - xcode-test-shard-calculation@%s:
+              inputs:
+              - shard_count: $TEST_SHARD_COUNT
+              - product_path: $BITRISE_XCTESTRUN_FILE_PATH
+          - deploy-to-bitrise-io@%s:
+              inputs:
+              - pipeline_intermediate_files: |-
+                  BITRISE_TEST_SHARDS_PATH
+                  BITRISE_TEST_BUNDLE_PATH
+                  BITRISE_XCTESTRUN_FILE_PATH
```
tothszabi, line 238: You do not need to expose the `BITRISE_XCTESTRUN_FILE_PATH` env var because the `BITRISE_TEST_BUNDLE_PATH` is enough.

_tests/integration/ios_test.go
```diff
@@ -204,6 +250,13 @@ configs:
               - test_repetition_mode: retry_on_failure
               - cache_level: none
           - deploy-to-bitrise-io@%s: {}
+        test_without_building:
+          steps:
+          - pull-intermediate-files@%s: {}
+          - xcode-test-without-building@%s:
+              inputs:
```
tothszabi, line 257: The `xctestrun` input is missing from the list. We need to set it to `$BITRISE_TEST_BUNDLE_PATH/[name of test run].xctestrun`. The tricky bit is that the `xctestrun` filename is constructed based on project values and stack settings. I think setting it to a non-existing file could enough here.

scanners/ios/workflow.go
```diff
@@ -15,6 +16,17 @@ const (
 	AutomaticCodeSigningValue             = "api-key"
 	CacheLevelKey                         = "cache_level"
 	CacheLevelNone                        = "none"
+	ShardCoundKey                         = "shard_count"
+	ShardCoundValue                       = "$TEST_SHARD_COUNT"
```
tothszabi, line 20: You have a spelling mistake in the name: `Cound` -> `Count`

scanners/ios/workflow.go
```diff
@@ -15,6 +16,17 @@ const (
...
+	ProductPathKey                        = "product_path"
+	ProductPathValue                      = "$BITRISE_XCTESTRUN_FILE_PATH"
+	PipelineIntermediateFilesKey          = "pipeline_intermediate_files"
+	PipelineIntermediateFilesValue        = "BITRISE_TEST_SHARDS_PATH\nBITRISE_TEST_BUNDLE_PATH\nBITRISE_XCTESTRUN_FILE_PATH"
+	OnlyTestingKey                        = "only_testing"
+	OnlyTestingValue                      = "$BITRISE_TEST_SHARDS_PATH/$BITRISE_IO_PARALLEL_INDEX"
+
+	// test pipeline
+	testPipelineID = "run_tests_parallel"
```
tothszabi, line 29: Do we need the `parallel` word in the name? The main goal of the pipeline is to run the tests and feels like `parallel` is just an implementation detail.

_tests/integration/fastlane_test.go
AndrasEszes: Let's ask Jeremy

_tests/integration/ios_test.go
AndrasEszes: Add a placeholder value for this input now and later we'll re-think how should resolve the correct `xctestrun` path.

scanners/ios/workflow.go
AndrasEszes: Rename it to `run_tests`.

tothszabi (approved):

PR #281, opened 2025-03-21
nodejs-scanner

zoltanszabo-bitrise:
### Version

Requires a *MAJOR/MINOR/PATCH* [version update](https://semver.org/)

### DIFF
--- a/_tests/integration/helper/validation.go
+++ b/_tests/integration/helper/validation.go
@@ -35,15 +35,31 @@ func ValidateConfigExpectation(t *testing.T, ID, expected, actual string, versio
 		fmt.Println("Expected: ", expPth)
 		fmt.Println("Actual: ", actPth)
 
-		_, err = exec.LookPath("opendiff")
-		if err == nil {
-			require.NoError(t, exec.Command("opendiff", expPth, actPth).Start())
+		tool, arguments, err := diffTool([]string{expPth, actPth})
+		if err != nil {
+			log.Warnf("unable to open config diff %s", err)
 			t.FailNow()
 			return
 		}
-		log.Warnf("opendiff not installed, unable to open config diff")
+
+		require.NoError(t, exec.Command(tool, arguments...).Start())
 		t.FailNow()
 	}
+}
+
+func diffTool(arguments []string) (string, []string, error) {
+	if _, err := exec.LookPath("code"); err == nil {
+		args := []string{"--diff"}
+		args = append(args, arguments...)
+
+		return "code", args, nil
+	}
+
+	if _, err := exec.LookPath("opendiff"); err == nil {
+		return "opendiff", arguments, nil
+	}
+
+	return "", nil, fmt.Errorf("not found opendiff or code")
 }
 
 func replaceVersions(str string, versions ...interface{}) (string, error) {
--- a/_tests/integration/manual_config_test.go
+++ b/_tests/integration/manual_config_test.go
@@ -246,6 +246,28 @@ var customConfigVersions = []interface{}{
 	steps.CacheSaveSPMVersion,
 	steps.DeployToBitriseIoVersion,
 
+	// node-js
+	// npm
+	models.FormatVersion,
+	steps.ActivateSSHKeyVersion,
+	steps.GitCloneVersion,
+	steps.NvmVersion,
+	steps.CacheRestoreNPMVersion,
+	steps.NpmVersion,
+	steps.NpmVersion,
+	steps.NpmVersion,
+	steps.CacheSaveNPMVersion,
+	// yarn
+	models.FormatVersion,
+	steps.ActivateSSHKeyVersion,
+	steps.GitCloneVersion,
+	steps.NvmVersion,
+	steps.CacheRestoreNPMVersion,
+	steps.YarnVersion,
+	steps.YarnVersion,
+	steps.YarnVersion,
+	steps.CacheSaveNPMVersion,
+
 	// other
 	models.FormatVersion,
 	steps.ActivateSSHKeyVersion,
@@ -483,6 +505,28 @@ var customConfigResultYML = fmt.Sprintf(`options:
                 config: default-macos-config
               none:
                 config: default-macos-config
+  node-js:
+    title: Project Directory
+    summary: The directory containing the package.json file
+    env_key: NODEJS_PROJECT_DIR
+    type: user_input
+    value_map:
+      "":
+        title: Node Version
+        summary: The version of Node.js used in the project. Leave it empty to use
+          the latest Node version
+        env_key: NODEJS_VERSION
+        type: user_input_optional
+        value_map:
+          "":
+            title: Package Manager
+            summary: The package manager used in the project
+            type: selector
+            value_map:
+              npm:
+                config: default-node-js-npm-config
+              yarn:
+                config: default-node-js-yarn-config
   react-native:
     title: Is this an [Expo](https://expo.dev)-based React Native project?
     summary: |-
@@ -770,6 +814,7 @@ configs:
           - certificate-and-profile-installer@%s: {}
           - restore-npm-cache@%s: {}
           - npm@%s:
+              title: npm install
               inputs:
               - workdir: $CORDOVA_WORK_DIR
               - command: install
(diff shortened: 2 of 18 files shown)


_tests/integration/manual_config_test.go
```diff
@@ -449,6 +449,11 @@ var customConfigResultYML = fmt.Sprintf(`options:
                 config: default-macos-config
               none:
                 config: default-macos-config
+  node-js: 
+	title: Project Root
+	summary: The directory containing the package.json file
+	env_key: PACKAGE_JSON_DIR
```
tothszabi, line 455: This is a yaml string and it seems like these lines have more white spaces than the type. I think this might cause some issues when parsing it.

_tests/integration/helper/validation.go
```diff
@@ -35,14 +33,14 @@ func ValidateConfigExpectation(t *testing.T, ID, expected, actual string, versio
...
 
-		_, err = exec.LookPath("opendiff")
-		if err == nil {
-			require.NoError(t, exec.Command("opendiff", expPth, actPth).Start())
-			t.FailNow()
-			return
-		}
-		log.Warnf("opendiff not installed, unable to open config diff")
+		//_, err = exec.LookPath("opendiff")
+		//if err == nil {
```
tothszabi, line 37: I saw a few commented out code in this PR. Could you remove the unnecessary ones?

scanners/ionic/ionic.go
```diff
@@ -2,6 +2,7 @@ package ionic
 
 import (
 	"fmt"
+	"github.com/bitrise-io/bitrise-init/scanners/nodejs"
```
tothszabi, line 5: Usually we organise the imports into groups. The first group is for the stdlib imports and then comes the rest. You can set up Goland to automatically use `goimports` when you save a file.

scanners/nodejs/nodejs_test.go
tothszabi, line 1: I think this test was only used for development, right?

scanners/nodejs/nodejs.go
```diff
@@ -0,0 +1,183 @@
...
+
+func (scanner *Scanner) ExcludedScannerNames() []string {
+	return []string{}
+}
+
+// Options returns the options for the scanner
+func (scanner *Scanner) Options() (models.OptionNode, models.Warnings, models.Icons, error) {
+	projectRootOption := models.NewOption(projectRootDirInputTitle, projectRootDirInputSummary, projectRootDirInputEnvKey, models.TypeSelector)
+	nodeVersionOption := models.NewOption(nodeVersionInputTitle, nodeVersionInputSummary, nodeVersionInputEnvKey, models.TypeOptionalUserInput)
+	packageManagerOption := models.NewOption(packageManagerInputTitle, packageManagerInputSummary, packageManagerInputEnvKey, models.TypeSelector)
```
tothszabi, line 113: We ask for the node version and package manager from the user. But the scanner tries to find these values from the project itself. Should not we only ask for these if the scanner could not determine it?

scanners/nodejs/nodejs.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// DefaultOptions returns the default options for the scanner
+func (scanner *Scanner) DefaultOptions() models.OptionNode {
+	projectRootOption := models.NewOption(projectRootDirInputTitle, projectRootDirInputSummary, projectRootDirInputEnvKey, models.TypeUserInput)
+	nodeVersionOption := models.NewOption(nodeVersionInputTitle, nodeVersionInputSummary, nodeVersionInputEnvKey, models.TypeOptionalUserInput)
+	packageManagerOption := models.NewOption(packageManagerInputTitle, packageManagerInputSummary, packageManagerInputEnvKey, models.TypeSelector)
+
+	for _, pkgMgr := range pkgManagers {
+		configOption := models.NewConfigOption("default-node-js-"+pkgMgr.name+"-config", nil)
```
tothszabi, line 138: Can we reuse the `configName` utility function you created for the identifier creation here?

_tests/integration/helper/validation.go
zoltanszabo-bitrise: Sure

scanners/nodejs/nodejs_test.go
zoltanszabo-bitrise: Yes

_tests/integration/helper/validation.go
```diff
@@ -35,13 +33,13 @@ func ValidateConfigExpectation(t *testing.T, ID, expected, actual string, versio
 		fmt.Println("Expected: ", expPth)
 		fmt.Println("Actual: ", actPth)
 
-		_, err = exec.LookPath("opendiff")
+		//_, err = exec.LookPath("opendiff")
```
tothszabi, line 36: Could you restore these lines?

scanners/cordova/cordova.go
```diff
@@ -2,6 +2,7 @@ package cordova
 
 import (
 	"fmt"
+	"github.com/bitrise-io/bitrise-init/scanners/nodejs"
```
tothszabi, line 5: For some reason this import group is not formatted. Could you do it?

scanners/nodejs/nodejs_test.go
tothszabi, line 1: I think this test was only used for dev purposes, right?

scanners/nodejs/utiltity.go
```diff
@@ -0,0 +1,352 @@
...
+	log.TPrintf("Checking package.json")
+	packageJsonPath := filepath.Join(searchDir, packageJson)
+	if exists := utility.FileExists(packageJsonPath); exists {
+		log.TPrintf("- %s - found", packageJsonPath)
+		return filepath.Rel(searchDir, packageJsonPath)
+	}
+
+	log.TPrintf("- %s - not found", packageJsonPath)
+	return "", fmt.Errorf("package.json not found")
+}
```
tothszabi, line 39: This function is not used.

scanners/nodejs/utiltity.go
```diff
@@ -0,0 +1,352 @@
...
+
+var nodeVersionFiles = []string{".nvmrc", ".node-version", ".tool-versions"}
+
+func checkNodeVersion(searchDir string) nodeVersionResult {
+	log.TPrintf("Checking Node version")
+
+	result := nodeVersionResult{
+		version: "",
+		file:    "",
+	}
```
tothszabi, line 49: You are saving this into a local variable but it is not used in the function apart from returning it at the end. You can just simply do 
```
return nodeVersionResult{}
```
and Go will intitialise the variables with their default values (which is "" for strings).

scanners/nodejs/utiltity.go
```diff
@@ -0,0 +1,352 @@
...
+	return result
+}
+
+func getNodeVersionFromToolVersions(fileContent string) string {
+	lines := strings.Split(fileContent, "\n")
+	for _, line := range lines {
+		fields := strings.Fields(line)
+		if len(fields) >= 1 && fields[0] == "nodejs" {
+			return fields[1]
+		}
```
tothszabi, line 89: If the `.tool-versions` file contains only `nodejs` then this will crash. I guess that is not a correct file content format but still we should guard against it.

scanners/nodejs/utiltity.go
```diff
@@ -0,0 +1,352 @@
...
+			configs[configName(descriptor)] = config
+		}
+		for _, pkgMgr := range pkgManagers {
+			descriptor.pkgManager = pkgMgr.name
+			config, err := generateConfigBasedOn(descriptor, sshKeyActivation)
+			if err != nil {
+				return nil, err
+			}
+			configs[configName(descriptor)] = config
+		}
```
tothszabi, line 303: I think this for loop should be in an `else` branch.

zsolt-vicze:
The resulting YML will have the 1.x.x `restore-npm-cache` which is already outdated. Can we update it to version "2"?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/bitrise-io/bitrise-init
 
-go 1.22.0
+go 1.23.0
```
tothszabi, line 3: Could you downgrade this to 1.22.0 as that is the default Go version on some of our stacks. If someone wants to run this on an older stack then it will not compile out of the box.

zoltanszabo-bitrise: go mod tidy rewrotes it

tothszabi (approved):

PR #283, opened 2025-04-01
Kotlin Multiplatform scanner

godrei:
### Version

### Context

This PR adds a new Kotlin Multiplatform project type scanner.

**Project type detection**

The new scanner utilizes a newly introduced Grdale project detector. 
A Gradle project is detected by looking for a Gradle wrapper script (`gradlew`). 
If a wrapper script is found, the detector further analyzes the project to search for dependencies:
- collecting the Version Catalog file (`libs.versions.toml`)
- collecting Build scripts (`build.gradle`) [by included projects and all build scripts]

Other scanners can decide on the concrete project type (Android/Kotlin Multiplkatform/...) based on the dependencies.

**Deafult workflows**

The Kotlin Multiplatform project type scanner generates a single workflow `run_tests`, which includes the usual repository access steps and a Gradle Test step to execute the `tests` Gradle task.


direntry/direntry.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+)
+
+var ignoreDirs = []string{".git", ".github", ".gradle", ".idea", "build", ".kotlin", ".fleet", "CordovaLib", "node_modules"}
+
+type DirEntry struct {
```
tothszabi, line 12: I see you wrote your own recursive filesystem directory walking logic. Have you thought simply wrapping the default `filepath.Walk`? I think it can do the same thing as what you are doing here.

If you need to skip a dir then you can return a special error:
```
err = filepath.Walk(path/to/dir, func(path string, info fs.FileInfo, err error) error {
	if iNeedToSkipThis {
		return filepath.SkipDir
	}
	...
})
```

If you need to limit the depth then you can check the returned path for how many components it has:
```
err = filepath.Walk(path/to/dir, func(path string, info fs.FileInfo, err error) error {
	if strings.Count(path, string(os.PathSeparator)) > maxDepth {
		return filepath.SkipDir
	}
	...
})
```

And if you want to find a single thing and terminate the execution then you can return another special error:
```
err = filepath.Walk(path/to/dir, func(path string, info fs.FileInfo, err error) error {
	if iNeedToStop {
		return filepath.SkipAll
	}
	...
})
```

direntry/direntry.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func recursiveListEntries(rootDir string, parent *DirEntry, currentDepth, maxDepth uint) error {
+	if currentDepth >= maxDepth {
+		return nil
+	}
+
+	entries, err := os.ReadDir(parent.AbsPath)
+	if err != nil {
+		// TODO: log error
```
tothszabi, line 91: Is this still a todo?

direntry/direntry.go
godrei: We discussed keeping the current directory parsing logic because filepath.Walk would make it harder to build relationship between directories and its parent entries.

detectors/gradle/direntry/direntry.go
```diff
@@ -0,0 +1,156 @@
...
+	return &parent, nil
+}
+
+// FindFirstEntryByName returns the first entry (shortest file path) with the specified name and directory status.
+func (e DirEntry) FindFirstEntryByName(name string, isDir bool) *DirEntry {
+	return recursiveFindFirstEntryByName([]DirEntry{e}, name, isDir)
+}
+
+// FindDirectEntryByName returns the direct child entry with the specified name and directory status.
+func (e DirEntry) FindDirectEntryByName(name string, isDir bool) *DirEntry {
```
tothszabi, line 47: nit: I had to read the comment to fully understand what `FindDirectEntryByName` does. It is clear that it only tries to find the entry in the first child level. I do not have a good name suggestion yet but maybe `FindDirectChildEntryByName` could be better. I am not a 100% sure in this name so this comment is totally optional.

tothszabi (approved):

tothszabi (approved):

PR #287, opened 2025-04-23
Java scanner

godrei:
### Version

### Context

This PR introduces a generic Java project type scanner.
Gradle and Maven based Java projects are detected by searching for Gradle wrapper and Maven wrapper scripts, the generated workflows run the projects' tests.


_tests/integration/android_test.go
```diff
@@ -18,32 +18,28 @@ import (
 func TestAndroid(t *testing.T) {
 	var testCases = []helper.TestCase{
 		{
-			"sample-apps-android-sdk22",
```
godrei, line 21: I added structure field names to all the test cases because a new field (`RelativeSearchDir`) was introduced, and I wanted to avoid listing an additional empty parameter for all the test cases.

_tests/integration/helper/execution.go
```diff
@@ -12,30 +12,59 @@ import (
...
 type TestCase struct {
-	Name             string
-	RepoURL          string
-	Branch           string
-	ExpectedResult   string
-	ExpectedVersions []interface{}
+	Name              string
+	RepoURL           string
+	Branch            string
+	RelativeSearchDir string
```
godrei, line 18: A new `RelativeSearchDir` was introduced to be able to run tests on a monorepo.

_tests/integration/helper/execution.go
```diff
@@ -12,30 +12,59 @@ import (
...
+	ExpectedResult    string
+	ExpectedVersions  []interface{}
+}
+
+type testHelper struct {
+	repoCache map[string]string
```
godrei, line 24: Repo caching is added to be able to run tests on a monorepo more efficiently.

scanners/java/java.go
```diff
@@ -0,0 +1,266 @@
...
+		s.mavenProject = mavenProject
+		printMavenProject(*mavenProject)
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func (s *Scanner) ExcludedScannerNames() []string {
+	return []string{}
```
ofalvai, line 92: Don't we want to exclude the Android scanner here?

ofalvai (approved):

scanners/java/java.go
godrei: It isn't needed, because the Java scanner is the last scanner in the available scanner's list: https://github.com/bitrise-io/bitrise-init/pull/287/files#diff-3f9980967933bdddf5b469600cd6e088d8042c92f27963c0eeb68d2c83a78cf2R85.
Exclusion works in a way, that if a scanner detects a project, then it can exclude next scanners.

scanners/kmp/kmp.go
```diff
@@ -90,6 +91,10 @@ func (s *Scanner) DetectPlatform(searchDir string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
+	if gradleProject == nil {
```
godrei, line 94: This is a fix for the current KMP scanner, `gradle.ScanProject` can return nil error and nil gradleProject, if the necessary project files weren't found.

ofalvai (dismissed):

tothszabi (approved):

PR #304, opened 2026-04-02
Nextjs

godrei:
### Version

### Context

Extends the existing Node.js scanner with framework detection (Next.js / NestJS) and declarative Node version setup. The workflow generation logic is unchanged — all Node.js projects produce the same workflow shape regardless of framework. The detected framework and Node version are stored on the project and surfaced in logs and the generated `bitrise.yml`.

### DIFF
--- a/scanners/nodejs/utiltity.go
+++ b/scanners/nodejs/utiltity.go
@@ -214,7 +214,6 @@ type configDescriptor struct {
 	hasLint     bool
 	hasTest     bool
 	isDefault   bool
-	framework   string
 	nodeVersion string
 }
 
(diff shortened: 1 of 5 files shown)


scanners/nodejs/utiltity.go
```diff
@@ -76,13 +57,6 @@ func checkPackageScripts(packageJsonPath string) (checkScriptResult, error) {
 		result.scripts = append(result.scripts, name)
 	}
 
-	if slices.Contains(result.scripts, "build") {
```
godrei, line 79: This was unused

molnarm (approved):

PR #305, opened 2026-04-14
Flutter web

godrei:
### Version

### Context

**Flutter scanner — web project support:**

- Added web project detection (`hasWebProject` via `web/` directory presence), using the new `WebProjectDir()` method from `go-flutter` v0.1.2
- Config names now encode platform presence as individual flags (`-ios`, `-android`, `-web`) instead of a combined platform string
- Replaced `appPlatform()` / `configName()` methods with standalone `targetPlatformInputValueFor()` and `configNameFor()` functions
- Simplified `DefaultOptions()`: removed the intermediate platform selector; the single default project now covers all three platforms (iOS + Android + Web)
- Renamed workflows from `primary`/`deploy` to `run_tests`/`build_app`
- Updated `run_tests` workflow description: clarifies that `flutter-test` runs when a test directory is present, otherwise `flutter-analyze` runs
- Deleted `const.go`; workflow description strings moved inline into `flutter.go`

**Dependencies:**

- Bumped `go-flutter` from v0.1.1 to v0.1.2-0.20260413123617-d1289effff65, which adds the `WebProjectDir()` method
- Vendored the new `WebProjectDir()` method

**Cleanup:**

- Removed unused `FindFileInAppDir` function from `utility/utility.go`

**Tests:**

- Expanded Flutter integration tests to cover web-only and iOS + Android + Web combined scenarios
- Updated Flutter versions in Flutter integration tests
- Updated Node.js integration tests to reflect Node 22.13.0 (was 22.0.0) in the sample app


scanners/flutter/flutter.go
```diff
@@ -167,30 +162,25 @@ func (scanner *Scanner) Options() (models.OptionNode, models.Warnings, models.Ic
...
 	}
 
 	return *flutterProjectLocationOption, models.Warnings{}, nil, nil
 }
 
 var defaultProjects = []project{
-	{hasTest: true, hasAndroidProject: true, hasIosProject: true},
-	{hasTest: true, hasAndroidProject: false, hasIosProject: true},
-	{hasTest: true, hasAndroidProject: true, hasIosProject: false},
+	{hasTest: true, hasAndroidProject: true, hasIosProject: true, hasWebProject: true},
```
godrei, line 173: This change simplifies the default workflows. Instead of generating configs for some target platform combinations (but not for all!), now we only create one which fits the ios, android and web targets combination.

molnarm (approved):

PR #308, opened 2026-04-27
Python scanner

godrei:
### Version

### Context

Python projects (FastAPI, Django, Flask, plain pytest apps) had no automatic CI config generation in the "Add new app" flow. This PR adds a Python scanner that covers both the automatic scan path and the manual configuration path.

### Changes

- Add `scanners/python` implementing `ScannerInterface`.
- Detection (positive if any marker file is found, excluding `.git`, `node_modules`, `.venv`, `venv`, `__pycache__`):
  - Marker files: `requirements.txt`, `pyproject.toml`, `Pipfile`, `setup.py`. Each containing directory is treated as a separate project (monorepo support).
  - Package manager priority: `uv.lock`, then `poetry.lock`, then `requirements.txt`. If none of these are present, the user picks during setup and a config is generated for every option.
  - Python version: `.python-version`, then `.tool-versions`, then `pyproject.toml` `requires-python` (operators stripped). Set as `tools: { python: <version> }` when found.
  - Pytest: detected via `pytest.ini`, `conftest.py`, `[tool.pytest…]` in `pyproject.toml`, or a `pytest` entry in any of the supported requirements files.
  - Dev requirements: first match of `requirements-dev.txt`, `requirements-test.txt`, `requirements_dev.txt`, `requirements_test.txt`. Installed alongside `requirements.txt` in the pip flow.
  - Poetry `--no-root` requirement: detected from `pyproject.toml`. Plain `poetry install` is emitted when `package-mode = false` is set, when `[tool.poetry] packages = ...` is declared, or when a default package layout exists at `<name>/__init__.py` or `src/<name>/__init__.py`. Otherwise `poetry install --no-root` is used.
  - Framework (FastAPI, Django, Flask) is logged for scan visibility only; it does not currently affect the generated workflow.
- Generated `run_tests` workflow per package manager:
  - pip: `pip install --upgrade pip; pip install -r requirements.txt` (plus the dev requirements file when detected), test with `pytest`.
  - poetry: `pip install poetry; poetry install [--no-root]`, test with `poetry run pytest`. Poetry is bootstrapped via pip because it is not pre-installed on the Bitrise stacks.
  - uv: `pip install uv; uv sync`, test with `uv run pytest`. Same bootstrap reason as Poetry.
  - Each manager has its own cache key (lockfile hash) and cache paths: `~/.cache/pip`, `~/.cache/pypoetry`, `~/.cache/uv`.
- Manual config prompts for `PYTHON_PROJECT_DIR`, `PYTHON_VERSION`, and **Package Manager** (pip / poetry / uv); a `bitrise tools install python $PYTHON_VERSION` script step replaces the declarative `tools` block. The manual Poetry config defaults to `--no-root` because per-project detection is not possible at generation time.
- Register the scanner in `scanners.ProjectScanners()` after `ruby` (no `ExcludedScannerNames`).
- Tests:
  - Unit tests for `parsePyproject` in `scanners/python/detection_test.go`.
  - Integration tests in `_tests/integration/python_test.go` against `bitrise-io/python-samples` (pip, poetry, uv fixtures).
  - Manual config tests extended for the Python defaults.

### Investigation details

- Initial implementation supported only pip; poetry and uv were added incrementally as `packageManagers` is a single slice that drives both scan and default config branches.
- Poetry and uv are not pre-installed on the Bitrise stacks, so the install scripts bootstrap them with `pip install <tool>` before running their commands. This avoids requiring stack changes for the feature to ship.
- The Poetry `--no-root` decision is the only meaningful per-project variant beyond the manager itself. Detection mirrors Poetry's own package resolution order (`package-mode = false`, explicit `packages`, default layout), so the generated config matches what `poetry install` would actually do for that project.

### Decisions

- `DetectPlatform` only collects candidate project dirs; richer parsing (package manager, version, test runner, dev requirements, `--no-root`) runs in `Options()` per the per-scanner execution flow.
- Python does not exclude any other scanner (no overlap with mobile / cross-platform scanners).
- Dev requirements are installed only in the pip flow; poetry and uv handle dev groups through their own lockfiles, so no extra step is needed there.
- The manual Poetry config defaults to `--no-root` because (a) detection is not possible at generation time, (b) `--no-root` succeeds in both app and library cases (pytest discovers in-tree packages without installation), (c) users who need root install can edit the workflow.

### DIFF
--- a/scanners/python/detection.go
+++ b/scanners/python/detection.go
@@ -1,0 +1,240 @@
+package python
+
+import (
+	"path/filepath"
+	"strings"
+
+	"github.com/bitrise-io/bitrise-init/utility"
+	"github.com/bitrise-io/go-utils/fileutil"
+	"github.com/bitrise-io/go-utils/log"
+	"github.com/bitrise-io/go-utils/pathutil"
+)
+
+var markerFiles = []string{
+	"requirements.txt",
+	"pyproject.toml",
+	"Pipfile",
+	"setup.py",
+}
+
+// requirementsFiles are checked in order when looking for pytest as a dependency.
+var requirementsFiles = []string{
+	"requirements.txt",
+	"requirements-dev.txt",
+	"requirements-test.txt",
+	"requirements_dev.txt",
+	"requirements_test.txt",
+}
+
+func collectPythonProjectDirs(searchDir string) ([]string, error) {
+	fileList, err := pathutil.ListPathInDirSortedByComponents(searchDir, false)
+	if err != nil {
+		return nil, err
+	}
+
+	excludeFilters := []pathutil.FilterFunc{
+		pathutil.ComponentFilter(".git", false),
+		pathutil.ComponentFilter("node_modules", false),
+		pathutil.ComponentFilter(".venv", false),
+		pathutil.ComponentFilter("venv", false),
+		pathutil.ComponentFilter("__pycache__", false),
+	}
+
+	seenDirs := map[string]bool{}
+	var projectDirs []string
+
+	for _, markerFile := range markerFiles {
+		filters := append(excludeFilters, pathutil.BaseFilter(markerFile, true))
+		paths, err := pathutil.FilterPaths(fileList, filters...)
+		if err != nil {
+			return nil, err
+		}
+		for _, p := range paths {
+			dir := filepath.Dir(p)
+			if !seenDirs[dir] {
+				seenDirs[dir] = true
+				projectDirs = append(projectDirs, dir)
+			}
+		}
+	}
+
+	return projectDirs, nil
+}
+
+func detectPackageManager(projectDir string) string {
+	log.TPrintf("Checking package manager")
+
+	if utility.FileExists(filepath.Join(projectDir, "poetry.lock")) {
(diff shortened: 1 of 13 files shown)


scanners/python/detection.go
```diff
@@ -0,0 +1,368 @@
...
+var requirementsFiles = []string{
+	"requirements.txt",
+	"requirements-dev.txt",
+	"requirements-test.txt",
+	"requirements_dev.txt",
+	"requirements_test.txt",
+}
+
+type pyprojectInfo struct {
+	poetryPackageModeFalse bool
```
molnarm: nit: this will make reading code a bit confusing. The scanner name (`poetryNeedsNoRoot`) is a bit clearer, use that or flip the flag (`poetryPackageMode`).

molnarm (dismissed):

scanners/python/detection.go
godrei: I went with poetryPackageModeDisabled here: https://github.com/bitrise-io/bitrise-init/pull/308/commits/7ebfb81c0fc21c89b289a7b65050dab95e894920

molnarm (approved):

